PR #4, opened 2014-06-11
Fork log/syslog in order to support custom tags.

robxu9:
The current log/syslog package does not allow modification of the syslog
tag after a network connection is dialed. Normally this is okay, but since
logspout monitors several different processes and forwards them, we also
need to tag each process.

Additionally, tag and priority are public fields now, so if there's a need to
change the default tag and priority, it can be done.

Also, dialing every time we need to send something to syslog sucks.

Signed-off-by: Robert Xu <robxu9@gmail.com>

### DIFF
--- a/logspout.go
+++ b/logspout.go
@@ -2,13 +2,14 @@ package main
 
 import (
 	"fmt"
-	"github.com/robxu9/logspout/pkg/syslog"
 	"log"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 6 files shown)


progrium:
Wow, I never merged this? Can you rebase. :)

progrium:
We now only use the syslog package for constants, etc as we've basically reimplemented it for our purposes.

PR #30, opened 2014-11-25
Removed attach on restart event

Jojo-Jojovich:
Removed attach on restart event, because when a docker container is restarted, it doubles up the log entries.


progrium:
Can you rebase and drop the binary? Thanks! :)

Jojo-Jojovich:
Hello, I've rebased and dropped the binary.  :)  Thanks!

PR #36, opened 2015-01-10
Resolves the merge conflict on pull/20 (rfc5424)

dansowter:


dansowter:
@brycekahle awesome work mate - we use loggly, too.

PR #39, opened 2015-02-04
allow comma separated syslog targets on startup

Rob-Johnson:
This PR allows a user to pass a comma separated list of default syslog targets, rather than just one.


geowa4:
:+1: The Docker cluster I'm working on could really benefit from this.

PR #41, opened 2015-02-08
Add redis route target

bcicen:
This is a rebase of https://github.com/progrium/logspout/pull/24.
Streams output to redis, appending to a list named after the container. Ideas for improvement or future features:
- option to append all logs to a specific key?
- using pub/sub instead of a list

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,13 +14,19 @@ Logspout is a very small Docker container (14MB virtual, based on busybox), so y
 
 ## Using logspout
 
-#### Route all container output to remote syslog
+#### Route all container output to a remote
 
 The simplest way to use logspout is to just take all logs and ship to a remote syslog. Just pass a default syslog target URI as the command. Also, we always mount the Docker Unix socket with `-v` to `/tmp/docker.sock`:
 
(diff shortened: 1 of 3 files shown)


progrium:
So based on the previous PR on this, it seems like there is demand. But it would be good to do it in a way that is most common. Can I get some feedback from others that want Redis on whether this is a good approach?

konsti:
Redis Lists and Channels are most common. A perfect solution would support both options (configurable), but starting with lists is fine for now as well.
e.g. [logstash redis input](http://logstash.net/docs/1.4.2/inputs/redis#data_type)

progrium:
This is a great opportunity to try the module system in master that will let you build logspout again a custom redis adapter package living in a separate repository.

progrium:
Actually, now that I've seen that Redis is basically the most popular way to make logstash HA, I'd take this as a first-class adapter.

bcicen:
Great; I see there's been a bit of refactoring so I'll update my branch for this PR accordingly soon.

rtoma:
I have created an adapter for sending events to Redis in Logstash json layout.
Feedback on https://github.com/rtoma/logspout-redis-logstash is much appreciated, since it my first Golang project.

PR #49, opened 2015-03-02
rabbitmq support. 

nicocesar:
added  rabbitmq support.


progrium:
Cool. I'll ask you to rebase after the new Adapter interface is up and documented.

progrium:
Also it is very likely we won't pull this into the main repo since we won't be able to maintain it, but the new module system will let it live in a separate repo you own, and we can include it in our readme.

michaelshobbs:
@nicocesar let us know if you ended up creating your own module and we can include it in the 3rd party registry

PR #53, opened 2015-03-05
Send docker container hostname in log messages

danielrowles-wf:
Allow you to send the container 'hostname' rather than the hostname of the host/container in which logspout is running when using the RFC5424 streamer.

### DIFF
--- a/logspout.go
+++ b/logspout.go
@@ -76,6 +76,10 @@ func main() {
 
 				r.Target.StructuredData = v.Get("structuredData")
 				r.Target.AppendTag = v.Get("appendTag")
+
+				if v.Get("useContainerHostname") == "true" {
+					r.Target.UseContainerHostname = true
+				}
 			}
 			router.Router.Add(&r)
 		}
--- a/router/attacher.go
+++ b/router/attacher.go
@@ -69,6 +69,7 @@ func (m *AttachManager) attach(id string) {
 	container, err := m.client.InspectContainer(id)
 	assert(err, "attacher")
 	name := container.Name[1:]
+	hostname := container.Config.Hostname
 	success := make(chan struct{})
 	failure := make(chan error)
 	outrd, outwr := io.Pipe()
@@ -99,7 +100,7 @@ func (m *AttachManager) attach(id string) {
 	_, ok := <-success
 	if ok {
 		m.Lock()
-		m.attached[id] = NewLogPump(outrd, errrd, id, name)
+		m.attached[id] = NewLogPump(outrd, errrd, id, name, hostname)
 		m.Unlock()
 		success <- struct{}{}
 		m.send(&AttachEvent{ID: id, Name: name, Type: "attach"})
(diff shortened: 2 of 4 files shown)


progrium:
Can you explain this use case?

danielrowles-wf:
We have a central syslog server, and would like to split the logs up into a directory per container. At the moment, the hostname that's sent in the syslog message is that of the logspout container, not of the container that generated the logs.

progrium:
This is because I expected people to run the logspout container using the
hostname of the container host, because that's still a relevant piece of
information. Usually more relevant than container, but obviously you still
want to know which container, which is why we put the container name in
tags. This actually maps relatively close to how you would do logs
normally: hostname is the host, tag for each process/service.

My first reaction is that using the container hostname would be duplicating
information since the container name is already in tags.

Since the hostname is the field that is easiest to organize logs with, the
question becomes, what should be the primary way to organize logs ... by
host or by container. You're suggesting by container, but I'm sure plenty
of people would say host.

On Mon, Mar 9, 2015 at 4:39 AM, Dan Rowles <notifications@github.com> wrote:

> We have a central syslog server, and would like to split the logs up into
> a directory per container. At the moment, the hostname that's sent in the
> syslog message is that of the logspout container, not of the container that
> generated the logs.
>
> —
> <https://github.com/gliderlabs/logspout/pull/53#issuecomment-77823861>.
>



-- 
Jeff Lindsay
http://progrium.com

danielrowles-wf:
In our case, we definitely aren't interested in the host, as we want to be able to let something else (fleet) handle placement and migration of services for us. Container hostname (rather than name) would be the most useful thing for us, as that lets us easily pull together all of the logs for a given service. There are plenty of other ways we can do this, but neither container name nor hostname of machine on which container is running are particularly useful for our specific use case.

progrium:
That is a valuable data point. Expect to see better support for this in an
upcoming release.

On Mon, Mar 9, 2015 at 7:59 AM, Dan Rowles <notifications@github.com> wrote:

> In our case, we definitely aren't interested in the host, as we want to be
> able to let something else (fleet) handle placement and migration of
> services for us. Container hostname (rather than name) would be the most
> useful thing for us, as that lets us easily pull together all of the logs
> for a given service. There are plenty of other ways we can do this, but
> neither container name nor hostname of machine on which container is
> running are particularly useful for our specific use case.
>
> —
> <https://github.com/gliderlabs/logspout/pull/53#issuecomment-77848459>.
>



-- 
Jeff Lindsay
http://progrium.com

danielrowles-wf:
In that case, let me elaborate a little more. Container hostname is usually far more useful for us than name, because there are specific services (rabbitmq when clustering being the most obvious one) where the hostname matters. Additionally, each container knows its own hostname (so it probably appears in logs, performance metrics, etc) whereas only the host really knows or cares about the container name.

In our specific case, the most useful format for log messages from containers would be:
date container_hostname consul_service_name, as that would make log collation a breeze. I appreciate that's not ideal for an open-source project though. Perhaps you could consider using go templates to generate the log messages?

progrium:
Master now uses the container hostname by default, which you can change to the logspout hostname if you set `SYSLOG_HOSTNAME={{.Hostname}}` (default is {{.Container.Config.Hostname}})

PR #54, opened 2015-03-06
Forward logs depending on container environment variable

danielrowles-wf:
Conditionally forward container logs based on whether or not a "LOGSPOUT"
environment variable is set to "true"

### DIFF
--- a/adapters/syslog/syslog.go
+++ b/adapters/syslog/syslog.go
@@ -1,6 +1,7 @@
 package syslog
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"log"
@@ -76,7 +77,14 @@ type SyslogAdapter struct {
 
 func (a *SyslogAdapter) Stream(logstream chan *router.Message) {
 	for message := range logstream {
-		err := a.tmpl.Execute(a.conn, &SyslogMessage{message, a})
+		buf := new(bytes.Buffer)
+		err := a.tmpl.Execute(buf, &SyslogMessage{message, a})
+		if err != nil {
+			log.Println("syslog:", err)
+			a.route.Close()
+			return
+		}
+		_, err = a.conn.Write(buf.Bytes())
 		if err != nil {
 			log.Println("syslog:", err)
 			a.route.Close()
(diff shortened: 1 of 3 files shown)


progrium:
In your case, how do you decide what should be logged through logspout and what shouldn't?

danielrowles-wf:
Depends on whether the container already handles logging in some other way. Also, for example, our log collation services are themselves containers, we don't want to send them their own stdout. The idea behind this patch was to allow us to decide at deploy time what to do on a case-by-case basis.

progrium:
What do you think about a simpler implementation of always (not specific to route) looking for a `LOGSPOUT_IGNORE` variable on a container and ignoring if that is set?

danielrowles-wf:
Sounds good to me - I'll see if I can get that done this afternoon and send a new PR

progrium:
Your PR looked pretty good, but we just merged the new modules architecture into master. Would you mind re-implementing? Basically in the same place, but a lot of names of changed and files moved around.

adapters/syslog/syslog.go
```diff
@@ -76,7 +77,14 @@ type SyslogAdapter struct {
 
 func (a *SyslogAdapter) Stream(logstream chan *router.Message) {
 	for message := range logstream {
-		err := a.tmpl.Execute(a.conn, &SyslogMessage{message, a})
```
progrium: Hmm, how necessary is this?

danielrowles-wf: From local testing using rsyslog, it seems that the golang templates package doesn't do any internal buffering, and writes data to the socket as soon as it is ready. This seems to result in multiple udp packets being transmitted, which rsyslog interprets as separate (and malformed) log messages.

danielrowles-wf: More specifically, without this internal buffering, tcpdump shows me 39 UDP packets received from a short-lived container that outputs 3 lines of text, whereas with this fix applied, I only see 3 packets, and these are also processed correctly by rsyslog.

progrium: Ah, great catch. Thank you.

PR #35, opened 2015-01-03
Add support for streaming to Kafka.

derekelkins:
sarama is MIT licensed

PR #61, opened 2015-03-24
Kafka logspout adapter

dylanmei:
This is a cut of the Kafka adapter/module discussed in #35.

Some things to discuss:

- I've overloaded route "address" with multiple brokers separated by "," followed by a "/" and a Kafka topic. (i.e., broker1:9020,broker2:9020/topic) The topic is so important here that it doesn't seem to fit well as a route "option".
- There are tons of Kafka producer options that can be added. I've setup "compression codec" but not gone any further yet.
- `adapters/raw` has `RAW_FORMAT`, which is a go template, and `adapters/syslog` has `SYSLOG_FORMAT` which is very much not a go template. Wondering if `KAFKA_FORMAT` is actually appropriate.
- Master branch is completely unusable for me, not sure if it's just me yet. The magic of `make dev` is gone. Most of what is in this PR worked great before the 'adapters' change, but I can't expect a merge until things settle down.

Happy to make changes, move it, etc.

### DIFF
--- a/adapters/kafka/kafka.go
+++ b/adapters/kafka/kafka.go
@@ -35,10 +35,10 @@ func NewKafkaAdapter(route *router.Route) (router.LogAdapter, error) {
 	}
 
 	var tmpl *template.Template
-	if text := os.Getenv("KAFKA_FORMAT"); text != "" {
+	if text := os.Getenv("KAFKA_TEMPLATE"); text != "" {
 		tmpl, err = template.New("kafka").Parse(text)
 		if err != nil {
-			return nil, fmt.Errorf("couldn't create template from KAFKA_FORMAT: %v", err)
+			return nil, err
 		}
 	}
 
--- a/adapters/kafka/kafka.go
+++ b/adapters/kafka/kafka.go
@@ -22,7 +22,6 @@ type KafkaAdapter struct {
 	route    *router.Route
 	brokers  []string
 	topic    string
-	config   *sarama.Config
 	producer sarama.AsyncProducer
 	tmpl     *template.Template
 }
@@ -41,8 +40,7 @@ func NewKafkaAdapter(route *router.Route) (router.LogAdapter, error) {
 		}
 	}
 
-	config := buildConfig(route.Options)
-	producer, err := sarama.NewAsyncProducer(brokers, config)
+	producer, err := sarama.NewAsyncProducer(brokers, newConfig())
 	if err != nil {
 		return nil, fmt.Errorf("couldn't create Kafka producer: %v", err)
 	}
@@ -51,7 +49,6 @@ func NewKafkaAdapter(route *router.Route) (router.LogAdapter, error) {
 		route:    route,
 		brokers:  brokers,
 		topic:    topic,
-		config:   config,
 		producer: producer,
 		tmpl:     tmpl,
 	}, nil
@@ -71,7 +68,7 @@ func (a *KafkaAdapter) Stream(logstream chan *router.Message) {
 	}
 }
 
-func buildConfig(options map[string]string) *sarama.Config {
+func newConfig() *sarama.Config {
 	config := sarama.NewConfig()
 	config.ClientID = "logspout"
 	config.Producer.Return.Errors = false
@@ -79,7 +76,7 @@ func buildConfig(options map[string]string) *sarama.Config {
 	config.Producer.Flush.Frequency = 1 * time.Second
 	config.Producer.RequiredAcks = sarama.WaitForLocal
 
-	if opt := options["compression.codec"]; opt != "" {
+	if opt := os.Getenv("KAFKA_COMPRESSION_CODEC"); opt != "" {
 		switch opt {
 		case "gzip":
 			config.Producer.Compression = sarama.CompressionGZIP
--- a/adapters/kafka/kafka.go
+++ b/adapters/kafka/kafka.go
@@ -34,11 +34,6 @@ func NewKafkaAdapter(route *router.Route) (router.LogAdapter, error) {
 		return nil, err
 	}
 
-	config, err := buildConfig(route.Options)
-	if err != nil {
-		return nil, fmt.Errorf("couldn't prepare Kafka config from route options: %v", err)
-	}
-
 	var tmpl *template.Template
 	if text := os.Getenv("KAFKA_FORMAT"); text != "" {
 		tmpl, err = template.New("kafka").Parse(text)
@@ -47,6 +42,7 @@ func NewKafkaAdapter(route *router.Route) (router.LogAdapter, error) {
 		}
 	}
 
+	config := buildConfig(route.Options)
 	client, err := sarama.NewClient(brokers, config)
 	if err != nil {
 		return nil, fmt.Errorf("couldn't create Kafka client: %v", err)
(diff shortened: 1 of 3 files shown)


progrium:
Difference between raw and syslog format snuck up on me. Perhaps `RAW_FORMAT` should be renamed `RAW_TEMPLATE` is it was originally.

Can you be specific about master being unusable? It's been working for everybody else that's tried it with the exception of a few bugs that have been fixed pretty quickly. `make dev` works for me as I use it for development...

progrium:
It might be better to put the addresses as query params (multiple `address` keys). At least for encoding this in URI I think this is more appropriate. Doozerd did something similar. Then maybe the Route.Address becomes a slice with at least one element. Then I think Route should support an optional Path field for adapters to make use of (another example being Redis that might take a key name, or other MQs with topics).

progrium:
Do you think you could hand off the options dictionary wholesale somehow to Kafka to support all their options without hardcoding them?

dylanmei:
With gliderlabs/latest I get output at :8000/logs but with `make dev` I do not. I can keep trying to isolate this and create an issue.

adapters/kafka/kafka.go
```diff
@@ -0,0 +1,129 @@
...
+	var tmpl *template.Template
+	if text := os.Getenv("KAFKA_FORMAT"); text != "" {
+		tmpl, err = template.New("kafka").Parse(text)
+		if err != nil {
+			return nil, fmt.Errorf("couldn't create template from KAFKA_FORMAT: %v", err)
+		}
+	}
+
+	config := buildConfig(route.Options)
+	client, err := sarama.NewClient(brokers, config)
```
eapache: fyi, you no longer need to create/manage a `Client` at all unless you need it for something specific, you can simply call `NewAsyncProducer(brokers, config)`

progrium:
Do you mean specifically with `/logs` or with `/logs/<filter>`? I found and fixed a bug last night that broke the latter, but the former was working fine.

eapache:
You will need to either have a separate goroutine reading off the producer's `Errors()` channel, or you will have to disable `config.Producer.Return.Errors`. Otherwise the producer could hang trying to return an error.

dylanmei:
@progrium just /logs. I run `make dev` and get
 
    2015/03/24 13:46:27 loading and persisting routes in /mnt/routes
    2015/03/24 13:46:27 logspout dev serving http on :8000

Elsewhere I run `docker run learn/ping ping github.com`. In console, logspout sees this as 

    2015/03/24 13:47:03 event: 1203f24832b1 create
    2015/03/24 13:47:03 event: 1203f24832b1 start

but there's nothing at `/logs`.

progrium:
Just to confirm, you're not running learn/ping with `-t` option? Just because that would do it. Though this behavior is the same as :latest since it was not likely to attach when you're running interactive, but it could be parameterized. Anyway, it might not be the issue, I'll try to reproduce.

dylanmei:
@eapache Thanks, you are exactly right. I will pull off the errors as well. And I some connection timeout handling is due.  I appreciate you taking a look.

dylanmei:
@progrium I'll move my repro steps into a separate issue. I'd like to still discuss the Kafka adapter's route address and options.

dylanmei:
@progrium regarding route address being like host1,host2/topic

I like your idea of using a `path` attribute as Kafka `topic`

I'm unclear where you are going with multiple addresses and slices. Given I'm calling the routesapi, how does this change?

    {
        "adapter": "kafka",
        "filter_sources": ["stdout", "stderr"],
        "address": "broker1:9092,broker2:9092",
        "path": "/spark-logs"
    }

progrium:
not sure i like it but it would always be an array:
```
{
    "adapter": "kafka",
    "filter_sources": ["stdout", "stderr"],
    "address": ["broker1:9092", "broker2:9092"],
    "path": "/spark-logs"
}
```
...even with the common case of one address:
```
{
    "adapter": "syslog",
    "filter_sources": ["stdout", "stderr"],
    "address": ["syslogserver:9092"],
}
```
special casing the value just being a string when only one is present feels better, but would require the type to describe it as an empty interface with type assertions all over.

progrium:
alternatively, it would be less code and only be "odd" for the few adapters like this, if we have an empty Address field, and let the address query params fall into options as usual, and the adapter just uses them there directly.

dylanmei:
Okay, well I just want to make the case that it's super common to refer to multiple zookeepers as zk1,zk2,zk3 and it doesn't seem strange to me at all, here.

progrium:
when referring to multiple zookeepers, is it as a URI though? I guess as long as net.URL parses it properly, we can just do it that way... unless it doesn't and you were parsing manually?

dylanmei:
I was parsing manually. It's not a URI at Route.Address so...

dylanmei:
Just one example, out in the wild, of referring to a collection of servers and then a logical location at that site: https://github.com/deric/puppet-mesos

What is "address", really? ;)

progrium:
cool, i guess seeing it in the wild helps. let's just do that then.

dylanmei:
re: "Do you think you could hand off the options dictionary to Kafka wholesale somehow" 

The [Kafka producer config documentation](https://kafka.apache.org/08/configuration.html) refers to the Java client library, whereas Sarama is a Go version of the client library and is configured differently. So I started my approach assuming people would go to Kafka's docs for information and not Sarama. But you're right, I don't want to be responsible for owning this mapping. :(

Suggestions?

eapache:
Just take a "saramaConfig" option which maps to a complete `*sarama.Config`?

progrium:
Yeah unless we can get sarama to support a public API for key value properties (which I suspect they might internally), I'd say we should just point to sarama config docs as well as Kafka config docs.

eapache:
> a public API for key value properties (which I suspect they might internally)

Nope, we just read fields off the `sarama.Config` struct directly. We've never hit a case that actually required dynamic keys, so a static typed struct is much simpler and cleaner.

dylanmei:
For reference, the Sarama config: https://github.com/Shopify/sarama/blob/master/config.go#L6-L119

@eapache logspout's options are name/value pairs and are also meant to be url-friendly, so would be expressed as 

    {...
        "options": {
            "Sarama.Net.DialTimeout": "10s",
            "Sarama.Producer.MaxMessageBytes": 123,
            "Sarama.Producer.Flush.Frequency": "500ms",
        }
    }

In this form, I imagine we'd be building this config up with reflection when the adapter starts up.

dylanmei:
Noteworthy: the logspout syslog adapter side-steps the problem with `SYSLOG_xxx` environment variables.

But I prefer the options path as we drop route configurations on the hosts with confd templates.

progrium:
Environment options are intended for global adapter config where options are for configuration that might changer per adapter instance. Rendering route configuration files is not something I thought of, but that also suggests environment is better for more static configuration that might be part of a release (as configuration usually is, except a few cases, namely service discovery is treated dynamic separately)

dylanmei:
I understand what you're saying about env variables, that they apply globally (relative to docker host logspout is running on). So for Syslog it's probable that we'd want to use the same `SYSLOG_FORMAT` across routes.

I've thrown in `KAFKA_COMPRESSION_CODEC` as an env setting as it would probably (but not definitely) also be the same across routes/topics. I can't say that about other settings, however.

I'd be happy to see this as a core logspout module but Kafka is still evolving and not as ubiquitous as syslog. If we can't find the silver bullet I'm willing to maintain the separate module. I am in #gliderlabs in irc when my computer is not asleep.

dylanmei:
Closing this due to PR #77

PR #64, opened 2015-03-26
Add simple Logstash adapter

maxpersson:
This adds a simpler version of the Logstash UDP adapter created in github.com/gliderlabs/logspout.

### DIFF
--- a/adapters/logstash/logstash.go
+++ b/adapters/logstash/logstash.go
@@ -19,7 +19,7 @@ type LogstashAdapter struct {
 	route *router.Route
 }
 
-// NewLogstashAdapter creates a LogstashAdapter with UDP transport.
+// NewLogstashAdapter creates a LogstashAdapter with UDP as the default transport.
 func NewLogstashAdapter(route *router.Route) (router.LogAdapter, error) {
 	transport, found := router.AdapterTransports.Lookup(route.AdapterTransport("udp"))
 	if !found {
--- a/adapters/logstash/logstash.go
+++ b/adapters/logstash/logstash.go
@@ -5,17 +5,12 @@ import (
 	"errors"
 	"log"
 	"net"
-	"os"
 
 	"github.com/gliderlabs/logspout/router"
 )
 
-var hostname string
-
 func init() {
 	router.AdapterFactories.Register(NewLogstashAdapter, "logstash")
-
-	hostname, _ = os.Hostname()
 }
 
 // LogstashAdapter is an adapter that streams UPD JSON to Logstash.
@@ -47,8 +42,10 @@ func (a *LogstashAdapter) Stream(logstream chan *router.Message) {
 		msg := LogstashMessage{
 			Time:     m.Time.Unix(),
 			Message:  m.Data,
-			Hostname: hostname,
+			Name:     m.Container.Name,
+			ID:       m.Container.ID,
 			Image:    m.Container.Config.Image,
+			Hostname: m.Container.Config.Hostname,
 		}
 		js, err := json.Marshal(msg)
 		if err != nil {
@@ -68,6 +65,8 @@ func (a *LogstashAdapter) Stream(logstream chan *router.Message) {
 type LogstashMessage struct {
 	Time     int64  `json:"time"`
 	Message  string `json:"message"`
-	Hostname string `json:"hostname"`
-	Image    string `json:"image"`
+	Name     string `json:"docker.name"`
+	ID       string `json:"docker.id"`
+	Image    string `json:"docker.image"`
+	Hostname string `json:"docker.hostname"`
 }
--- a/adapters/logstash/logstash.go
+++ b/adapters/logstash/logstash.go
@@ -3,7 +3,7 @@ package logstash
 import (
 	"encoding/json"
 	"errors"
-	"io"
+	"log"
 	"net"
 	"os"
 
@@ -42,7 +42,7 @@ func NewLogstashAdapter(route *router.Route) (router.LogAdapter, error) {
 	}, nil
 }
 
-func (adapter *LogstashAdapter) Stream(logstream chan *router.Message) {
+func (a *LogstashAdapter) Stream(logstream chan *router.Message) {
 	for m := range logstream {
 		msg := LogstashMessage{
 			Time:     m.Time.Unix(),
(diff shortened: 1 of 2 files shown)


adapters/logstash/logstash.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"github.com/gliderlabs/logspout/router"
+)
+
+var hostname string
+
+func init() {
+	router.AdapterFactories.Register(NewLogstashAdapter, "logstash")
+
+	hostname, _ = os.Hostname()
```
progrium: What's this?

adapters/logstash/logstash.go
```diff
@@ -0,0 +1,63 @@
...
+
+// LogstashAdapter is an adapter that streams UPD JSON to Logstash.
+type LogstashAdapter struct {
+	conn  net.Conn
+	route *router.Route
+}
+
+// NewLogstashAdapter creates a LogstashAdapter with UDP transport.
+func NewLogstashAdapter(route *router.Route) (router.LogAdapter, error) {
+	transport, found := router.AdapterTransports.Lookup(route.AdapterTransport("udp"))
```
progrium: This is actually saying "udp" is the default transport (by using `route.AdapterTransport`), but allows any other supported transport if specified ("logstash+tcp://"). Does logstash support other transports or should this be hardcoded to udp?

adapters/logstash/logstash.go
```diff
@@ -0,0 +1,63 @@
...
+	}, nil
+}
+
+func (adapter *LogstashAdapter) Stream(logstream chan *router.Message) {
+	for m := range logstream {
+		msg := LogstashMessage{
+			Time:     m.Time.Unix(),
+			Message:  m.Data,
+			Hostname: hostname,
+			Image:    m.Container.Config.Image,
```
progrium: I'm guessing logstash just supports arbitrary JSON keys? If so, don't you think some people might want more than just container image? Do you have ideas on how to make this more configurable?

adapters/logstash/logstash.go
```diff
@@ -0,0 +1,63 @@
...
+		}
+		js, _ := json.Marshal(msg)
+		io.WriteString(adapter.conn, string(js))
```
progrium: You don't need to use `io.WriteString` when you're converting a byte array to a string... you can just say `adapter.conn.Write(js)`.. be sure to error check though.

adapters/logstash/logstash.go
maxpersson: I believe TCP should work also, but I have not tried it.

adapters/logstash/logstash.go
maxpersson: That is true. At this point we are doing further parsing of the message (which is in logfmt format by logrus) in the filter section of logstash. Maybe the container ID and some other Docker metadata could be good to have as keys.

adapters/logstash/logstash.go
progrium: Okay. Your comments suggest otherwise, so maybe consider removing specific mention of UDP in them.

adapters/logstash/logstash.go
progrium: I see, it's a global. Maybe set this up as a field of the adapter object and do this in its constructor.

adapters/logstash/logstash.go
progrium: It seems like this adapter is not as general as you think. Maybe it deserves to live as your own third-party module? 

I say this because if I were to do a general Logstash adapter, I probably wouldn't use JSON+UDP which is just one possible input format ... If I were to pick a canonical Logstash adapter machanism, I'd use the [logstash-forwarder](https://github.com/elastic/logstash-forwarder) protocol, perhaps even use it directly as a library. 

Don't you agree?

maxpersson:
@progrium Updated with more Docker fields, similar to the fields in https://github.com/digital-wonderland/docker-logstash-forwarder.

progrium:
Why do you think JSON+UDP should be the canonical way logspout should send logs to logstash?

maxpersson:
I think UPD should be the default because it's faster for streaming logs, and there can be a lot from some docker containers. For more info: https://github.com/dwbutler/logstash-logger#what-type-of-logger-should-i-use

progrium:
Maybe it should just be called a JSON adapter then?

progrium:
Does this satisfy your use?
https://github.com/gliderlabs/logspout/pull/65

maxpersson:
I think your JSON adapter is a pretty cool implementation, but I would like to have both! My implementation is focused on performance as this is more important in our use case.

progrium:
I recommend putting this adapter in its own repo and building logspout
against it. Others can use it if they want but you can continue to maintain
it.

On Friday, March 27, 2015, Max Persson <notifications@github.com> wrote:

> I think your JSON adapter is a pretty cool implementation, but I would
> like to have both! My implementation is focused on performance as this is
> more important in our use case.
>
> —
> <https://github.com/gliderlabs/logspout/pull/64#issuecomment-86850831>.
>


-- 
Jeff Lindsay
http://progrium.com

maxpersson:
Sure, are there any guidelines on how to do that?

progrium:
The repo name should be prefixed with `logspout-` and have a descriptive name, so I imagine this would be `logspout-logstash-json` on your github account. The go package can still be called `logstash` or anything, it's not relevant. You can put the package anywhere in there, but I recommend at the root. This way, you can just add a line to `modules.go` imports:
```
    _ "github.com/maxpersson/logspout-logstash-json"
```

maxpersson:
Will this be added to gliderlabs/logspout/modules.go or to my own fork?

progrium:
Anybody's fork. We'll list it as a third-party module in the readme.

maxpersson:
What I meant was when using logspout; will I be able to run your container with my module referenced in it, or would I create and push my own container from my own fork using my adapter?

progrium:
For now you need your own container fork. Eventually you can dynamically load modules at container boot.

maxpersson:
Ok, thanks for the info.

progrium:
This might make it easier for you: https://github.com/gliderlabs/logspout/pull/66/files

maxpersson:
I created a custom module instead, available here https://github.com/looplab/logspout-logstash. Closing this PR.

PR #65, opened 2015-03-27
json adapter

progrium:
This is a generic JSON adapter that *can* be used with Logstash, but doesn't need to be. Its fields can be configured with environment variables, with minimal default fields enabled, and a few useful premade fields that can be enabled.


adapters/json/json.go
```diff
@@ -0,0 +1,139 @@
...
+
+var configDefaults = map[string]string{
+	"JSON_FIELDS":          "time:uint,message,docker.hostname,docker.image",
+	"JSON_TIME":            "{{.Time.Unix}}",
+	"JSON_MESSAGE":         "{{.Data}}",
+	"JSON_SOURCE":          "{{.Source}}",
+	"JSON_DOCKER_HOSTNAME": "{{.Container.Config.Hostname}}",
+	"JSON_DOCKER_IMAGE":    "{{.Container.Config.Image}}",
+	"JSON_DOCKER_ID":       "{{.Container.ID}}",
+	"JSON_DOCKER_NAME":     "{{.ContainerName}}",
```
maxpersson: Missing ".", should be "{{.Container.Name}}".

adapters/json/json.go
```diff
@@ -0,0 +1,139 @@
...
+	"encoding/json"
+	"errors"
+	"log"
+	"net"
+	"os"
+	"strconv"
+	"strings"
+	"text/template"
+
+	"github.com/gliderlabs/logspout/adapters/syslog"
```
maxpersson: Depending on another adapter should be avoided IMO.

adapters/syslog/syslog.go
```diff
@@ -81,13 +81,13 @@ type SyslogAdapter struct {
...
 	for message := range logstream {
-		m := &SyslogMessage{message, a.conn}
+		m := NewSyslogMessage(message, a.conn)
 		buf, err := m.Render(a.tmpl)
 		if err != nil {
 			log.Println("syslog:", err)
 			return
 		}
-		_, err = a.conn.Write(buf)
+		_, err = a.conn.Write(buf.Bytes())
```
maxpersson: Why is there fixes to the syslog adapter in this PR?

adapters/json/json.go
progrium: This is intentional.
https://github.com/gliderlabs/logspout/blob/master/adapters/syslog/syslog.go#L135

adapters/json/json.go
progrium: I completely disagree.

adapters/syslog/syslog.go
progrium: It's not a fix, it's part of a change required by this PR that helps generalize syslog.

adapters/json/json.go
```diff
@@ -0,0 +1,139 @@
...
+		conn:  conn,
+		tmpl:  tmpl,
+		types: types,
+	}, nil
+}
+
+func (a *JSONAdapter) Stream(logstream chan *router.Message) {
+	defer a.route.Close()
+	for message := range logstream {
+		m := syslog.NewSyslogMessage(message, a.conn)
```
maxpersson: This still seems unnecessary to me, why create an instance of the syslog message just to render a template? It makes the code so much harder to read, especially when passing the connection. It's not idiomatic Go.

progrium: First, what is your suggested alternative. Second, it's sort of core to being able to specify new JSON fields.

maxpersson: I suggest using template.Execute() directly here and remove the dependency on syslog, or make the syslog message some kind of base message to use in several modules.

adapters/json/json.go
maxpersson: I think separation of packages is of highest priority for a maintainable system, but as noted below it can be solved by moving/renaming the syslog message to some kind of base message where it's clear that it's supposed to be used by several adapters.

adapters/json/json.go
progrium: You understand that the dependency on the syslog adapter is no different than the dependency on the core router package, right? They are both core packages to the logspout project.

adapters/json/json.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+		_, err = a.conn.Write(data)
+		if err != nil {
+			log.Println("json:", err)
+			return
+		}
+	}
+}
+
+func buildMap(input string, types map[string]string) map[string]interface{} {
```
maxpersson: It concerns me that the template rendering and value parsing happens on every log message. I think the goal of any forwarder is to be lightweight and performant, and leave the parsing to the receiver.

adapters/json/json.go
maxpersson: Of course, this is more about semantics and code readability.

adapters/json/json.go
progrium: I agree. Perhaps you can optimize this while maintaining configurable properties. For many this will be just fine. I encourage you to benchmark and show at what point this will not be viable to use.

maxpersson: Sure, I may do a benchmark if time allows, there will most likely be a big difference. For some people configurability matters more, for some raw performance.

progrium: Thank you for re-stating assumptions.

adapters/json/json.go
progrium: Thank you for the suggestion.

adapters/json/json.go
maxpersson: Sorry, it wasn't meant to be offending.

progrium: No I understand. I was getting caught up in style of communication. I appreciate your feedback and contributions, and any benchmarking around this will be very helpful to document. Not necessarily to prove a difference, but to show what scenario this adapter should not be considered for.

maxpersson: Cool, I'll let you know if I make a benchmark.

NoumanSaleem:
@progrium any update on this?

tumd:
I would really love to see this feature merged.

geoffreytran:
Another +1 on getting this merged.

dkiser:
@progrium Any way we can get `map` types properly converted to json before this PR is merged?  I tried running this branch on a dev box to see what `{{.Container.Config.Labels}}` would look like, but it doesn't appear the types map properly reads maps to end up being json marshalled correctly.

**example json pretty print of a syslog message with json adapter and labels field**
```json
{
  "docker": {
    "hostname": "6965fdb0f180",
    "image": "centos:centos7",
    "labels": "map[name:CentOS Base Image vendor:CentOS build-date:2015-12-23 license:GPLv2]"
  },
  "message": "test",
  "time": "1453494436"
}
```

**Ran with following env vars**
```bash
JSON_FIELDS="time:uint,message,docker.hostname,docker.image,docker.labels"
JSON_DOCKER_LABELS="{{.Container.Config.Labels}}"
```

Thoughts?

progrium:
That looks like an easy fix. Also I'm surprised this was never merged. Maybe it would be a good idea to merge this and then talk about fixing that as a separate issue that others might be able to help with?

dkiser:
@progrium either way is fine.  

Any ideas how the marshalling would be fixed?  Looking at golang json docs seems like it would be marshalled correctly by default for known types.  Is the use of the template render causing this?  Haven't looked too deeply at it plus I'm a go beginner.

progrium:
Yeah I'm assuming it's some oversight where I assume some variable is going to be a string. I haven't looked at the code in a while. I'm just, you know, swamped. Also still looking for another maintainer for this project.

progrium:
Ugh, and it needs to be rebased now too. Maybe neeeext week? :\

dkiser:
Hehe No rush on my part.  I can always build and use locally with the changes 8-).

dkiser:
@progrium  one thing I noticed was this code is using the stock udp transport.  I had a specific need to send JSON over a valid syslog message.  So i'll possibly fork and re-use some of this json code by adding a new MODE to the syslog adaptor configured by environment variables.  The idea would be to send JSON payload in the "message" portion of a syslog message where this payload is configured via environment variables as you've already started on this branch.

High level goal is JSON payload + syslog (udp or tcp).

progrium:
No reason adapters can't use each other. If there was a way to refactor this so that syslog could be configured to use it (and all its JSON configuration) in its messages that would be rad. Try and explore that in your fork and if you come up with anything let us know!

PR #66, opened 2015-03-27
allowing easy custom builds of logspout

progrium:
ONBUILD hook for overriding `modules.go` and example directory


maxpersson:
This looks fantastic, a really slim solution for adding functionality! Thanks for adding.

PR #70, opened 2015-03-31
Add ONBUILD to build custom modules, fixes #69

### LINKED ISSUE #69: Container build does not compile custom modules
The `RUN` command in the Dockerfile (https://github.com/gliderlabs/logspout/blob/master/Dockerfile#L8) needs to be prefixed with `ONBUILD` to actually run custom module builds. I'll submit a PR that fixes this.

maxpersson:
This should make the custom module functionality work as intended.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ EXPOSE 8000
 
 COPY . /go/src/github.com/gliderlabs/logspout
 ONBUILD COPY ./modules.go /go/src/github.com/gliderlabs/logspout/modules.go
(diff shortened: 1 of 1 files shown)


progrium:
This breaks normal logspout though...

progrium:
This is actually more involved than I thought. We also need to keep the source around.

progrium:
https://github.com/gliderlabs/logspout/pull/73

maxpersson:
Will be fixed in #73, closing.

PR #71, opened 2015-03-31
Add URL for Logstash module

maxpersson:
I'm not sure which format you want for custom modul links, this is a suggestion. What do you think is appropriate?


progrium:
Looks good!

PR #73, opened 2015-03-31
custom builds for real this time

progrium:


maxpersson:
Tested and works as intended. LGTM. Thanks for fixing!

PR #82, opened 2015-04-11
Update README run instructions

shazow:
- progrium/logsprout -> gliderlabs/logsprout.
- Volume-mount the correct path for the docker socket.
- Publish port on 127.0.0.1 rather than publicly.
- Assign a name to the container in examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,8 @@ You can also download and load a specific version:
 
 The simplest way to use logspout is to just take all logs and ship to a remote syslog. Just pass a syslog URI (or several comma separated URIs) as the command. Also, we always mount the Docker Unix socket with `-v` to `/var/run/docker.sock`:
 
-	$ docker run \
-		-v=/var/run/docker.sock:/var/run/docker.sock \
+	$ docker run --name="logsprout" \
+		--volume=/var/run/docker.sock:/tmp/docker.sock \
 		gliderlabs/logspout \
 		syslog://logs.papertrailapp.com:55555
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,8 +24,8 @@ You can also download and load a specific version:
 
 The simplest way to use logspout is to just take all logs and ship to a remote syslog. Just pass a syslog URI (or several comma separated URIs) as the command. Also, we always mount the Docker Unix socket with `-v` to `/var/run/docker.sock`:
 
-	$ docker run \
-		-v=/var/run/docker.sock:/var/run/docker.sock \
+	$ docker run --name="logsprout" \
```
shazow: Should this have `-d` too?

README.md
```diff
@@ -43,10 +43,11 @@ You can tell logspout to ignore specific containers by setting an environment va
 
 Using the [httpstream module](http://github.com/gliderlabs/logspout/blob/master/httpstream), you can connect with curl to see your local aggregated logs in realtime. You can do this without setting up a route URI.
 
-	$ docker run -d -p 8000:8000 \
-		-v=/var/run/docker.sock:/var/run/docker.sock \
-		progrium/logspout
-	$ curl $(docker port `docker ps -lq` 8000)/logs
```
shazow: If you want to keep the dynamic ip lookup, then we probably want `--expose` rather than `--publish`?

progrium: I don't understand "the dynamic ip lookup". 

This line makes `--expose` on the CLI unnecessary:
https://github.com/gliderlabs/logspout/blob/master/Dockerfile#L4

Unless you have access to the Docker bridge, you need to use `--publish`.

shazow: If we do something like `--publish=127.0.0.1:8000:8000` (to avoid having the port be externally available), then we can just hard-code 127.0.0.1 as before rather than look it up using ```docker port `docker ps -lq` 8000```.

progrium: Ah, right. Or even `localhost` for easier typing.

shazow: Yea, except `--publish` fails with localhost (expects an ip), at least on the version I tried.

progrium: Right. But I mean, when you use it. You're talking about examples.

shazow: So you want to publish on 127.0.0.1 but curl localhost? I think consistency should win over ease of typing, especially since this will be mostly for reading purposes and possibly copypasta.

progrium: You're right.

shazow: <3

rtrauntvein:
:+1:

Issue #75, opened 2015-04-07
panic: runtime error: assignment to entry in nil map


nagliyvred:
Hi @progrium,

When I run the current master pointing to loggly url like that: `rfc5424://logs-01.loggly.com:514?structuredData=<token>@41058`
I get an exception like that: 
```
2015-04-07T00:48:16.268748415Z # logspout v3-master by gliderlabs
2015-04-07T00:48:16.268825080Z # adapters: tcp syslog raw udp
2015-04-07T00:48:16.268952603Z # options : persist:/mnt/routes
2015-04-07T00:48:16.269574474Z panic: runtime error: assignment to entry in nil map
2015-04-07T00:48:16.269611719Z 
2015-04-07T00:48:16.269611719Z goroutine 16 [running]:
2015-04-07T00:48:16.269611719Z runtime.panic(0x7af000, 0xa1d5d3)
2015-04-07T00:48:16.269611719Z 	/usr/lib/go/src/pkg/runtime/panic.c:279 +0xf5
2015-04-07T00:48:16.269611719Z github.com/gliderlabs/logspout/router.(*RouteManager).AddFromUri(0xc208024de0, 0x7fff0d7f8d38, 0x5a, 0x0, 0x0)
2015-04-07T00:48:16.269611719Z 	/go/src/github.com/gliderlabs/logspout/router/routes.go:102 +0x4b3
2015-04-07T00:48:16.269611719Z github.com/gliderlabs/logspout/router.(*RouteManager).Setup(0xc208024de0, 0x0, 0x0)
2015-04-07T00:48:16.269611719Z 	/go/src/github.com/gliderlabs/logspout/router/routes.go:184 +0x1bb
2015-04-07T00:48:16.269611719Z main.main()
2015-04-07T00:48:16.269611719Z 	/go/src/github.com/gliderlabs/logspout/logspout.go:39 +0x646
2015-04-07T00:48:16.269611719Z 
2015-04-07T00:48:16.269611719Z goroutine 17 [runnable]:
2015-04-07T00:48:16.269611719Z runtime.MHeap_Scavenger()
2015-04-07T00:48:16.269611719Z 	/usr/lib/go/src/pkg/runtime/mheap.c:507
2015-04-07T00:48:16.269611719Z runtime.goexit()
2015-04-07T00:48:16.269611719Z 	/usr/lib/go/src/pkg/runtime/proc.c:1445
2015-04-07T00:48:16.269611719Z 
[...]

Issue #56, opened 2015-03-08
Support HTTP streamer for Sumo Logic


harlow:
Sumo Logic has a "Hosted Collector" but its via HTTP and not Syslog:
https://service.sumologic.com/help/Setting_up_a_Hosted_Collector.htm

If its something that Logspout team would consider as PR for the project I wouldn't mind writing the code for it (just need a little direction):

We have syslog streamers defined here:
https://github.com/gliderlabs/logspout/blob/master/router/streamers.go#L43-L54

However, then there is an Exported `HttpStreamer` but that seems to be writing to HTTP Response object and not a remote HTTP endpoint.
https://github.com/gliderlabs/logspout/blob/master/router/streamers.go#L116

A new `httpStreamer` function might be a bit confusing (as far as naming is concerned). Any suggestions?

progrium:
We're about to merge the `adapter` branch into master which has a new API for creating what were previously called "streamers". If you work against that branch now, the easier it will be to pull it in. Depending on specific or uncommon its use might be, we may recommend to have it live as a separate module repo.

harlow:
@progrium glad I asked! Will make sure to work off of that branch. Thanks

progrium:
Adapters now merged into master

raychaser:
hi there. i am actually with Sumo Logic, and i had a rough and tumble implementation pre-refactored Logspout. @progrium you might remember me emailing you about this. just getting back into some hacking with the goal of making an HTTP adapter and noticed this open issue. @harlow do you have any code already? happy to discuss live in #gliderlabs.

progrium:
Ping me in IRC if anybody needs help working with what's in master. It
should be straightforward looking at other adapters, but definitely want to
learn any pain points, etc.

On Thu, Apr 9, 2015 at 7:56 PM, raychaser <notifications@github.com> wrote:

> hi there. i am actually with Sumo Logic, and i had a rough and tumble
> implementation pre-refactored Logspout. @progrium
> <https://github.com/progrium> you might remember me emailing you about
> this. just getting back into some hacking with the goal of making an HTTP
> adapter and noticed this open issue. @harlow <https://github.com/harlow>
> do you have any code already? happy to discuss live in #gliderlabs.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/gliderlabs/logspout/issues/56#issuecomment-91391669>.
>



-- 
Jeff Lindsay
http://progrium.com

andyshinn:
[...]

PR #87, opened 2015-04-14
HTTP transport with configurable buffering

raychaser:
Among other things, in response to #56.

This adds an adapter that sends logs to an HTTP(S) endpoint via POST. To prevent round-trip madness, the adapter buffers mesages and sends them only when the buffer is full (default buffer capacity: 100 messages), or a configurable timeout has been reached (default: 1 second). I took some care to make sure keep-alive is honored, so this should be reasonably efficient.

Tested against Sumo Logic with a good amount of concurrent load, but this should work against any HTTP-ish endpoint. Of course, this is very much an initial implementation that likely can be much improved.

`adapters/http/README.md` has some more details, and an initial TODO list.

This includes the fix for #75 from #76.

Also, I am not at all sure whether I am doing the right thing with the VERSION file!

@progrium, there's an "Issue found" section in the README that I would be happy to discuss either as part of the PR, or on IRC (i am `raychas3r`, usually in PDT)

@dylanmei, FYI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,8 @@ dev:
 		-v /var/run/docker.sock:/var/run/docker.sock \
 		-v $(PWD):/go/src/github.com/gliderlabs/logspout \
 		-p 8000:80 \
-		-e ROUTE_URIS=$(ROUTE) \
+		-e ROUTE_URIS="$(ROUTE)" \
+		-e LOGSPOUT=ignore \
 		$(NAME):dev
 
 build:
--- a/VERSION
+++ b/VERSION
@@ -1,1 +1,1 @@
-v3-master
+v3-raychaser-http-buffered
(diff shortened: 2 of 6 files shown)


raychaser:
so, i finally figured out how to use the custom build option to actually pick up my own adapter after staring at @dylanmei's Kafka adapter for a long time - the trick is that the source has to be on Github. given this, i think it might be wise to continue to harden this adapter over at https://github.com/raychaser/logspout-http.

progrium:
At a glance this looks pretty good but it might be a bit before I can really take a look. With your repo up people can try it in the meantime.

raychaser:
yup, i know it is not a small change. i am going to close the PR for now. 

any chance you could merge #76? it would make my life a bit easier!

progrium:
Hah, yes. Sorry I didn't merge that sooner.

PR #89, opened 2015-04-15
Support TLS connections to Docker

josharian:
Respect DOCKER_TLS_VERIFY and DOCKER_CERT_PATH
environment variables. I believe that boot2docker
sets these up by default.

### DIFF
--- a/router/pump.go
+++ b/router/pump.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"log"
 	"os"
+	"path/filepath"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


progrium:
There is nothing wrong with this, but ... you shouldn't run logspout on a different host and if it's on the same host, is TLS necessary?

josharian:
Fair points.

Consider this to be support for development, not production. I bumped into this while playing with logspout. It was easier to add support in logspout than to fight with boot2docker to figure out how to turn off TLS. I suspect I'm not the only one.

progrium:
Oh, so running logspout on os x hitting the boot2docker VM? Did you have a reason not to use the dev container? (`make dev`)

josharian:
Yes, that. No reason, I just didn't notice the dev container.

However:

```bash
$ make dev
flag provided but not defined: -f

Usage: docker build [OPTIONS] PATH | URL | -

Build a new image from the source code at PATH

  --force-rm=false     Always remove intermediate containers, even after unsuccessful builds
  --no-cache=false     Do not use cache when building the image
  -q, --quiet=false    Suppress the verbose output generated by the containers
  --rm=true            Remove intermediate containers after a successful build
  -t, --tag=""         Repository name (and optionally a tag) to be applied to the resulting image in case of success
make: *** [dev] Error 2
```

josharian:
Hmm, maybe I have an old Docker client version. Updating...

josharian:
Updated. `make dev` works.

I think I'd still rather have `go run logspout.go` work, though. Up to you.

progrium:
Try using `make dev` for a bit and let me know if you still think that.

josharian:
Aye aye. Will let this idle and report back. Please consider it abandoned if I've not said anything either way by June 1, 2015. (I may not spend more time on logspout immediately.)

inhumantsar:
+1 to this. Our devs need it for docker-machine.

michaelshobbs:
Closing for now since the use case is dev-related and you can always use `make dev` instead.

PR #97, opened 2015-05-01
Fix port number in httpstream example

fw42:
The internal port is [80](https://github.com/gliderlabs/logspout/blob/2bd141a23632fce6aeca99f98bf7d51c3956612f/router/http.go#L10), not 8000.


fw42:
ping @progrium if you have a minute. thanks.

PR #98, opened 2015-05-05
fixed publish arg, internal http port is 8000

alex-sherwin:


michaelshobbs:
:+1: this definitely got in the way of getting started :smile:

PR #100, opened 2015-05-11
Adds a .Env function on syslog messages

dfuentes:
I know the usual method to do changes like this is rolling a custom logspout container with our modified syslog adapter module, but I thought this change might be generally useful.

We've run into the issue where we want to add some container specific information to our log lines, but don't have control of either the container hostname or container name (we are on mesos/marathon).  Allowing using container environment variables in the syslog configuration templates (i.e. SYSLOG_TAG) allows us to pull in this custom info.

### DIFF
--- a/adapters/syslog/syslog.go
+++ b/adapters/syslog/syslog.go
@@ -8,6 +8,7 @@ import (
 	"log/syslog"
 	"net"
 	"os"
+	"strings"
 	"text/template"
 	"time"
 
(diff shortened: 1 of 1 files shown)


ejholmes:
:+1: :+1: this would be really handy.

sebest:
Any feedback on adding this feature? Would be really useful on ECS, thanx

sebest:
I found a solution using labels on ECS and the SYSLOG_TAG environment variable of logspout:

SYSLOG_TAG="{{ index .Container.Config.Labels \"com.amazonaws.ecs.container-name\" }}"

Carles-Figuerola:
Look at #182, it should solve this problem and can be changed at runtime to the desired .Env key you want

dfuentes:
Closing this, it's pretty old at this point, and I no longer use logspout

PR #103, opened 2015-05-29
allow to force-set black background of httpstream output

t0mk:
On white or bright-grey terminal, it's very difficult to read the color of httpstream ouput. I added an optional query param backgroud=black which will force-set the terminal background of the httpstream output text to black.


progrium:
wouldn't you prefer just turning colors off?

t0mk:
I know it's poss to turn the colors off, but I think the colors are really useful to distinguish output of different containers. I don't want to lose the readability just because I prefer white-background terminal.

I was thinking how to implement it least-invasively and this was what I came up with. I am happy to rewrite it if you have another idea how to keep the colors in bright terminal.

PR #104, opened 2015-05-29
Allow populating cached routes

brendangibat:
I had an instance where I wanted to fill out multiple routes for a container dynamically and didn't want a new container repo every time and didn't want to mount a volume for configuration to it. This solved the problem for it's startup - we've since moved away from this solution but I wanted to send the PR back to the mainline to see if it was wanted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ The simplest way to use logspout is to just take all logs and ship to a remote s
 		gliderlabs/logspout \
 		syslog://logs.papertrailapp.com:55555
 
-logspout will gather logs from other containers that are started **without the `-t` option**. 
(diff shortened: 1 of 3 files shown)


progrium:
Cool. This doesn't seem all that different then specifying multiple URI arguments to create routes on boot. What was the thinking behind this approach besides no mounts?

brendangibat:
Well I needed to specify specific tags and syslog format for each of the endpoint URI's, and I didn't see a way to get those associations lined up with multiple arguments.

progrium:
Perfect, thanks.

brendangibat:
Hey I'm going to close this out since I'm not using it anymore and I don't think there's a need for it.

PR #109, opened 2015-06-05
Add functionality for configuring syslog facility

banjiewen:
Abusing the `user` facility wasn't an option in my infrastructure. This has some limited (preprod) testing and seems to work just fine.

I'm pretty much completely new to Go, so let me know if there are any stylistic changes necessary.

### DIFF
--- a/adapters/syslog/syslog.go
+++ b/adapters/syslog/syslog.go
@@ -16,9 +16,11 @@ import (
 )
 
 var hostname string
+var facility syslog.Priority
 
 func init() {
 	hostname, _ = os.Hostname()
+	facility = get_facility()
 	router.AdapterFactories.Register(NewSyslogAdapter, "syslog")
 }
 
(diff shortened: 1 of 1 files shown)


progrium:
Great idea and good first pass. As an improvement, I'd do the mapping with a package variable using a `[]string` using the constants (which are ints) as indexes (it looks like a map):
https://github.com/golang/glog/blob/master/glog.go#L110-Lundefined

And then the lookup is much simpler:
https://github.com/golang/glog/blob/master/glog.go#L154

This will probably end up being much more compact.

josegonzalez:
@banjiewen mind updating your PR to implement the ideas mentioned by @progrium ?

banjiewen:
@josegonzalez: I haven't been using this project for a while, so I'm probably not going to finish that off, sorry. Time is limited :\

josegonzalez:
Okay, no problem. Sorry it took forever to follow up.

PR #112, opened 2015-06-09
update the forked repo

ymote:


mgood:
@ymote is this a change you were making for yourself, or something you'd like to contribute to the main repository? If this is a change you would like to contribute, please update the Pull Request title and description to summarize the intent of these changes. This new shell script only works with syslog, and changes the way you would start the Docker container, so it would need some changes, but if we better understood your intent we could help find a way to better support it.

PR #113, opened 2015-06-11
Update example streaming usage with correct port mapping

bcicen:
As currently logspout listens on port 8000.


zucaritask:
+1. It took me a bit to realize there was a bug there.

bcicen:
fixed in #140

PR #114, opened 2015-06-15
Allow Docker label to be used to ignore containers.

cloudnautique:
This change allows a user to set a Docker label `com.gliderlabs.logspout.ignore=true` to have Logspout ignore the container for log processing. I took the liberty of proposing the namespace based on Docker's documentation. If there is something else that should be used please let me know. I didn't see any previous label support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,9 +39,9 @@ You can tell logspout to ignore specific containers by setting an environment va
 
         $ docker run -d -e 'LOGSPOUT=ignore' image
 
-Alternatively, can use the Docker label 'com.gliderlabs.logspout.ignore=true' to ignore containers.
+Alternatively, can use the Docker label 'logspout.ignore=true' to ignore containers.
 
-        $ docker run -d -l 'com.gliderlabs.logspout.ignore=true' image
+        $ docker run -d -l 'logspout.ignore=true' image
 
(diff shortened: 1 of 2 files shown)


progrium:
Where is Docker documentation on namespaces?

cloudnautique:
The docs are here:
https://docs.docker.com/userguide/labels-custom-metadata/

progrium:
I've not decided on how to best use labels yet, but the article does say:
> Keys without namespace (dots) are reserved for CLI use. This allows end- users to add metadata to their containers and images without having to type cumbersome namespaces on the command-line.

Which suggests since our labels are for end-user metadata use, we don't need to namespace. Perhaps a compromise is `logspout.ignore` or I might suggest a more general `logging.ignore` ... thoughts?

cloudnautique:
This seemed like an easy place to introduce a label. My other thoughts on label usage, without regard to implementation at this point, would be for routing.

I personally like `logspout.ignore` because it is specific to where the label applies. My expectation with the more general `logging.ignore` would be for all logging tools to ignore it. In some cases, I might like another tool to handle logging for a particular container.

I do like the shorter namespace. It potentially impacts future label usage so I will defer to you. Let me know and I'll make the change. Do you prefer commits squashed?

PR #126, opened 2015-09-02
Fix port to publish

glogiotatidis:
logspout listens by default on port 8000


glogiotatidis:
My bad, I was using latest instead of master

PR #127, opened 2015-09-02
Add TLS transport. Fixes #116.

### LINKED ISSUE #116: TLS Support
I noticed logspout doesn't have tls support and was considering submitting an implementation. I had two questions.

- Are there any reasons why using logspout with tls might be a bad idea? 
- Are you aware of any tricky issues that need to be tackled in order to implement this?

Thanks,
Shanon

glogiotatidis:
Tested against papertrail and works as expected. You can try with `giorgos/logspout` docker image.


progrium:
Nice.

Issue #107, opened 2015-06-02
Broken Pipe when Sending TCP Data to Loggly


brianwallace:
I am using logspout to stream docker logs into Loggly over TCP.  This is done using the `syslog+tcp://logs-01.loggly.com` endpoint:

```
    /usr/bin/docker run --name logspout \
    -v=/var/run/docker.sock:/tmp/docker.sock \
    -e SYSLOG_STRUCTURED_DATA=[LOGGLY_CUSTOMER_TOKEN]@41058  \
    gliderlabs/logspout:master \
    syslog+tcp://logs-01.loggly.com:514
```

Way too frequently, however, I am seeing a broken pipe message and data streaming from logspout stops.  From the logspout logs:

`2015/06/02 01:04:07 syslog: write tcp 54.236.68.122:514: broken pipe`

I am assuming this occurs when the [Loggly Route 53 ingestion endpoint changes IP](https://www.loggly.com/blog/why-aws-route-53-over-elastic-load-balancing/) and the TCP connection is lost.

Logspout should be updated to attempt to reestablish this connection if it is lost.

rlister:
This is definitely not a loggly-specific problem. It seems the new adaptor code causes this problem both for TCP and UDP (surprisingly). I am writing to a local syslog endpoint: with `v2` or `latest` (i.e. release branch) I can write UDP, stop the endpoint, restart it, and logspout carries on writing. With `master`, using either TCP or UDP, if I restart the endpoint the connection is dropped and not recreated.

brianwallace:
@rlister, thanks for the feedback.  I have updated to logspout `v2` and will give this a shot.  I should be able to confirm in a couple days whether or not this solves my issue (it sounds like it will!).

fastmap22:
Any updates on this. I am getting this error too with (raw+tcp)

logspout v3-master by gliderlabs

raw: write tcp <ip>:17000: broken pipe

brianwallace:
FYI, I was not able to go back to a previous release `v2` or `latest` (i.e. release branch) because it does not support the new endpoint format (to the best of my knowledge):

for example:
`syslog+tcp://logs-01.loggly.com:514`

fastmap22:
Any updates on this?

strzelecki-maciek:
would love to see this fixed as well. Currently logspout is not trying to reconnect after a broken pipe err, which effectively makes it very unreliable

arabold:
[...]

PR #150, opened 2015-11-29
Syslog: Re-establish TCP/TLS transport on connection interruptions

janeczku:
This PR fixes issue #107 where logspout indefinitely stops sending syslog messages when the connection is interrupted (e.g. the syslog consumer's IP address changed) and has to be manually restarted.

This establishes a new connection when a syslog write operation on a TCP/TLS transport adapter fails with a network related temporary error.

### DIFF
--- a/adapters/syslog/syslog.go
+++ b/adapters/syslog/syslog.go
@@ -9,6 +9,8 @@ import (
 	"net"
 	"os"
 	"reflect"
+	"sync"
+	"syscall"
 	"text/template"
 	"time"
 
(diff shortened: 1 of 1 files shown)


adapters/syslog/syslog.go
```diff
@@ -91,7 +115,9 @@ func (a *SyslogAdapter) Stream(logstream chan *router.Message) {
 		if err != nil {
 			log.Println("syslog:", err)
 			if reflect.TypeOf(a.conn).String() != "*net.UDPConn" {
-				return
+				if ok := a.maybeReconnect(err); !ok {
+					return
```
keis: Should there be a second attempt and sending the data here?

janeczku: @keis i don't know if there "should" be.  With UDP transport failed sends aren't retried either. Anyhow, i'd rather first hear back from @progrium whether he is intending to go forward with this.

Issue #136, opened 2015-10-27
Connection Reset By Peer


jgallen23:
I'm running exactly how the readme says:

```
15:18 $ docker run -d --name="logspout" \
>     --volume=/var/run/docker.sock:/tmp/docker.sock \
>     --publish=127.0.0.1:8000:80 \
>     gliderlabs/logspout
664111310c88bb3f010dd9f887507de4dc908f2bc8d81c61bc57debc8c3a2dba
```

and then I get connection reset by peer:

```
15:18 $ curl http://127.0.0.1:8000/logs
curl: (56) Recv failure: Connection reset by peer
```

What other information can I provide to help debug?

martinhynar:
I faced exactly the same problem. Logspout inside container is attached to port 8000 so `--publish=127.0.0.1:8000:8000` should do the trick.

PR #166, opened 2016-03-03
v3

### LINKED ISSUE #116: TLS Support

mattaitchison:
### Fixed
- use start/die like old version not create/destroy
- performance fix, generalizing SyslogMessage, minor cleanups
- Initialize Route options map
- Fixed a couple of typos, updated narrative
- UDP message delivery should not kill the program
- Exit with return code 1 on job setup failure
- Simplify and add early exit to RoutingFrom
- Unmarshal without buffering
- Remove unnecessary closure
- Undo change introduced in 07555c5
- Fix port number in httpstream example
- Use correct nilvalue for structured data as per rfc 5424
- retry tcp errors and don't hang forever on failure

### Added
- mention irc channel
- allowing easy custom builds of logspout
- Allow env vars in stream URLs
- Allow you to ignore log messages from individual containers by setting container environment variable, LOGSPOUT=ignore, when starting
- Add URL for Logstash module
- Adding CircleCI, Docker and IRC badges to readme.
- Add TLS transport. Fixes #116

### Removed
- Removed attach on restart event
- remove dev containers
- Removed deprecated library hosted in google code in favor of its new home

### Changed
- switched to gliderlabs org
- assume build
- rough pass at breaking logspout.go into separate packages
- fully split up packages. major refactoring of router
- simpler matching. working routesapi. dropped old utils
- make sure all uri params get into route options
- readme updates and module specific readmes
- renamed ConnectionFactory to AdapterTransport
- updated readme to use current schema
- names and parama
- more readable
- hold handler from returning until streamer finishes
- primarily designed new boot output, but came with it architectural changes
- updating docker sock location
- support old location for docker socket
- force link in case its run again, such as with custom builds
- analytics test
- update analytics
- Update README.md
- Update README with tls module
- Wrong port in README.md #136


michaelshobbs:
:+1:

Issue #159, opened 2016-02-12
Exclude containers by label (feature)


sshipway:
I am using Logspout to capture output on all hosts in our Rancher cluster.  Certaing container stacks need to be excluded, but it is not as easy to add an env var as it is to add a label to a previously defined service stack.

It would be very useful to be able to exclude a container by setting a container label, rather than by using the `LOGSPOUT=exclude` environment variable.  If the name of the label could be configured via an env var on the logspout container, then this would be even better.

Also, if you could specify by *inclusion* instead it would be helpful.

EG, 
`docker run ... -e EXCLUDE=logspout.exclude -e INCLUDE=logspout.include gliderlabs/logspout`

Then containers with the label `logspout.include` would be include, unless they also had the label `logspout.exclude`.  Of course, if not defined, `EXCLUDE` would default to nothing, and `INCLUDE` would default to everything.

ebr:
+1 for this. I'm also using Rancher, and there's currently an issue where cadvisor spams the logs, so as a workaround i'd like to exclude it from logspout temporarily. If implemented as suggested by @sshipway, I'd just add a `--label logspout.exclude= ` to my `docker run` command for the Rancher agent.

PR #171, opened 2016-03-24
Exclude containers by label.

selimekizoglu:
This allows to ignore containers with a particular label specified by:
`docker run -e EXCLUDE=logspout.exclude gliderlabs/logspout`

This will ignore all the containers ran using `--label logspout.exclude=...` as suggested (#159)

### DIFF
--- a/router/pump.go
+++ b/router/pump.go
@@ -60,6 +60,11 @@ func ignoreContainer(container *docker.Container) bool {
 			return true
 		}
 	}
+	excludeLabel := getopt("EXCLUDE", "")
+	_, ok := container.Config.Labels[excludeLabel]
+	if ok {
+		return true
+	}
 	return false
 }
 
--- a/router/pump.go
+++ b/router/pump.go
@@ -62,7 +62,7 @@ func ignoreContainer(container *docker.Container) bool {
 	}
 	excludeLabel := getopt("EXCLUDE_LABEL", "")
 	if value, ok := container.Config.Labels[excludeLabel]; ok {
-		return strings.ToLower(value) == "true"
+		return len(excludeLabel) > 0 && strings.ToLower(value) == "true"
 	}
 	return false
 }
--- a/router/pump.go
+++ b/router/pump.go
@@ -60,10 +60,9 @@ func ignoreContainer(container *docker.Container) bool {
 			return true
 		}
 	}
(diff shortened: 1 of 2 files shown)


josegonzalez:
This needs documentation, mind providing that?

josegonzalez:
Also duplicates #114.

selimekizoglu:
@josegonzalez Sorry. Updated README.md.

router/pump.go
```diff
@@ -60,6 +60,11 @@ func ignoreContainer(container *docker.Container) bool {
 			return true
 		}
 	}
+	excludeLabel := getopt("EXCLUDE", "")
```
josegonzalez: Maybe `EXCLUDE_LABEL` would be better?

router/pump.go
```diff
@@ -60,6 +60,11 @@ func ignoreContainer(container *docker.Container) bool {
...
+	_, ok := container.Config.Labels[excludeLabel]
+	if ok {
```
josegonzalez: Can we check that the value is a string matching `true`?

router/pump.go
```diff
@@ -60,10 +60,9 @@ func ignoreContainer(container *docker.Container) bool {
 			return true
 		}
 	}
-	excludeLabel := getopt("EXCLUDE", "")
-	_, ok := container.Config.Labels[excludeLabel]
-	if ok {
-		return true
+	excludeLabel := getopt("EXCLUDE_LABEL", "")
+	if value, ok := container.Config.Labels[excludeLabel]; ok {
```
josegonzalez: What happens when `excludeLabel` is an empty string? Just curious.

josegonzalez:
Sorry I'm really bad at review :(

router/pump.go
selimekizoglu: If `EXCLUDE_LABEL` is an empty string then containers with a label that has an empty key get ignored:

`$ docker run -d --label =true image`

josegonzalez: Sounds weird. Can we not do that and check that the label is greater than zero in length?

selimekizoglu: Makes sense.

josegonzalez:
Looks good, I assume you've tested this latest iteration?

selimekizoglu:
Yes, with an empty and a non-empty environment variable.

PR #180, opened 2016-04-19
update container name if we get a rename event. closes #144

### LINKED ISSUE #144: Detect container rename
logspout does not detect the new container name after it is renamed:
```
compassionate_knuth:  web.1    | info: HTTP GET / url=/, connection=upgrade, host=boer.host.com, x-forwarded-proto=http, x-forwarded-for=10.10.10.10, x-forwarded-port=80, x-request-start=1447862791.577, accept-encoding=gzip, deflate, accept=*/*, user-agent=HTTPie/0.8.0, method=GET, httpVersion=1.1, originalUrl=/, , statusCode=404, responseTime=4
```
We are using dokku and noticed that stoping/starting logspout resolves this issue.

```
~# dokku logspout:stop
04e262444bb7ee254af13cd01d9a5e4e2f79c329eed35c35d2262b8aad2d040b
~# dokku logspout:start
logspout
```
```
logspout:  2015/11/18 16:07:12 loading and persisting routes in /mnt/routes 
logspout:  2015/11/18 16:07:12 logspout v2 serving http on :8000 
logspout:  2015/11/18 16:07:12 routing all to syslog://logs.papertrailapp.com:XXXX
...
boer.web.1:  web.1    | info: HTTP GET / url=/, connection=upgrade, host=boer.host.com, x-forwarded-proto=http, x-forwarded-for=10.10.10.10, x-forwarded-port=80, x-request-start=1447862841.258, accept-encoding=gzip, deflate, accept=*/*, user-agent=HTTPie/0.8.0, method=GET, httpVersion=1.1, originalUrl=/, , statusCode=404, responseTime=2
```
But it would be nice if logspout could automatically detect a container renaming.

michaelshobbs:
Uses new container name if we receive a `rename` event in the docker event stream

thanks to @josegonzalez for the code here
closes #144

### DIFF
--- a/router/pump.go
+++ b/router/pump.go
@@ -92,10 +92,9 @@ func (p *LogsPump) Setup() error {
 func (p *LogsPump) rename(event *docker.APIEvents) {
     p.mu.Lock()
     defer p.mu.Unlock()
-    debug("pump.rename() event.Actor.ID:", event.Actor.ID)
-    container, err := p.client.InspectContainer(event.Actor.ID)
+    container, err := p.client.InspectContainer(event.ID)
     assert(err, "pump")
-    pump, _ := p.pumps[normalID(event.Actor.ID)]
+    pump, _ := p.pumps[normalID(event.ID)]
     pump.container.Name = container.Name
 }
 
@@ -116,7 +115,7 @@ func (p *LogsPump) Run() error {
 		return err
 	}
 	for event := range events {
-		debug("pump.Run() event:", normalID(event.Actor.ID), event.Status)
+		debug("pump.Run() event:", normalID(event.ID), event.Status)
 		switch event.Status {
 		case "start", "restart":
 			go p.pumpLogs(event, true)
@@ -130,8 +129,6 @@ func (p *LogsPump) Run() error {
 }
 
 func (p *LogsPump) pumpLogs(event *docker.APIEvents, backlog bool) {
-	debug("pump.pumpLogs() event.ID:", event.ID)
-	debug("pump.pumpLogs() event.Actor.ID:", event.Actor.ID)
 	id := normalID(event.ID)
 	container, err := p.client.InspectContainer(id)
 	assert(err, "pump")
@@ -180,8 +177,7 @@ func (p *LogsPump) pumpLogs(event *docker.APIEvents, backlog bool) {
 func (p *LogsPump) update(event *docker.APIEvents) {
 	p.mu.Lock()
 	defer p.mu.Unlock()
-	debug("pump.update() event.Actor.ID:", event.Actor.ID)
-	pump, pumping := p.pumps[normalID(event.Actor.ID)]
+	pump, pumping := p.pumps[normalID(event.ID)]
 	if pumping {
 		for r, _ := range p.routes {
 			select {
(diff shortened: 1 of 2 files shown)


router/pump.go
```diff
@@ -89,6 +89,15 @@ func (p *LogsPump) Setup() error {
 	return nil
 }
 
+func (p *LogsPump) rename(event *docker.APIEvents) {
+    p.mu.Lock()
+    defer p.mu.Unlock()
+    container, err := p.client.InspectContainer(event.ID)
```
mattaitchison: Seems that `event.ID` has been removed in 1.22. Although [go-dockerclient](https://github.com/fsouza/go-dockerclient/blob/master/event.go#L39-L42) is maintaining backward compatability, I think it would be best to use `event.Actor.ID` in new code. What do you think?

https://github.com/fsouza/go-dockerclient/blob/master/event.go#L39-L42

michaelshobbs: We totally could do that. There are a few other spots that use `event.ID` as well. The question is does logspout want to require docker `>= 1.10.0` at this time or should we wait another couple minor API revs?

mattaitchison: Logspout wouldn't require docker `>= 1.10.0` since go-dockerclient maintains forward and backward compatibility.

michaelshobbs: Ah I misread that then. Well if there's no compatibility issues then I'm happy to update this PR.

router/pump.go
```diff
@@ -118,15 +130,17 @@ func (p *LogsPump) Run() error {
 }
 
 func (p *LogsPump) pumpLogs(event *docker.APIEvents, backlog bool) {
+	debug("pump.pumpLogs() event.ID:", event.ID)
+	debug("pump.pumpLogs() event.Actor.ID:", event.Actor.ID)
 	id := normalID(event.ID)
```
michaelshobbs: using `event.Actor.ID` here caused a panic at startup. 
Additional debugging shows `event.Actor.ID` is null for some reason in the startup case
`ef13d311a6c3` is the logspout container
```
2016/04/20 18:59:51 pump.pumpLogs() event.ID: ef13d311a6c3
2016/04/20 18:59:51 pump.pumpLogs() event.Actor.ID:
2016/04/20 18:59:51 pump: json: cannot unmarshal array into Go value of type docker.Container
```

michaelshobbs:
@mattaitchison updated

router/pump.go
mattaitchison: I guess we should use `event.ID`. Since `pump.Update` will also get a null `event.Actor.ID`.

```
2016/04/21 01:19:32 pump.pumpLogs() event.ID: cedf1a9662b1
2016/04/21 01:19:32 pump.pumpLogs() event.Actor.ID:
2016/04/21 01:19:32 pump.update() event.Actor.ID:
2016/04/21 01:19:32 pump.pumpLogs(): cedf1a9662b1 started
2016/04/21 01:19:32 pump.pumpLogs() event.ID: b2b5390b3129
2016/04/21 01:19:32 pump.pumpLogs() event.Actor.ID:
2016/04/21 01:19:32 pump.update() event.Actor.ID:
2016/04/21 01:19:32 pump.pumpLogs(): b2b5390b3129 started
```

Issue #163, opened 2016-03-01
Custom tagging/source on logs sent to Syslog (feature)


sshipway:
We have a number of Docker hosts that may be sending logs to Syslog via logspout.  These logs appear in syslog with the Docker host IP as the source.  However, we are only interested in the cluster source, not the specific host.

It would be useful to have an optional env var that can be set when starting the Logspout container, whereby you can specify the Source to use in the Syslog feed (and/or a tag to prepend to the syslog entry).  In this way, the logs would arrive at syslog with a common Source of (eg) 'cluster A' rather than one of the various IP addresses of the cluster members.

Also, being able to use the Container name or ID to prepend the log entry would help in analysis where the central log server gets thousands of such logs a second.

As Docker infrastructures grow in size, this could be come a very useful feature to have available.

alackmann:
Would love to see this as well. The Docker Cloud service (formerly Tutum) seems to set the `host` as the container name, which when you're logging staging/production into Loggly, results in the same hostname and an inability to tell environements apart.

sshipway:
I've been exploring the code; although it's written in GO and there's little to no documentation, it is possible that setting the SYSLOG_HOSTNAME environment variable will do this.  I'm running some tests.  If it does, then SYSLOG_PRIORITY, SYSLOG_TAG,  and possible SYSLOG_STRUCTURED_DATA will be able to work as well...

alackmann:
[tutum/syslogger](https://github.com/tutumcloud/syslogger) suggests this can be done - but I've tried passing the host in as an environment variable and it doesn't seem to take effect.

But I agree - it should be feasible to pass in tags using a similar method.

sshipway:
Tests seem to show that SYSLOG_TAG defaults to the container name (which is good) and SYSLOG_HOSTNAME defaults to the container ID (which is bad).  It does not affect the syslog source as sent by the Syslog protocol, but it does change the syslog message:
````
1457064896 2016 Mar  4 17:14:56 +13:00 10.10.25.249: 1: 2016-03-04T04:14:56Z Rancher: rancher-agent 30049 - [] time="2016-03-04T04:14:56Z" level="info" msg="Assigning IP [10.42.47.213/16], ContainerId [81f1b90ff0bce730f1104a28d28db07a5a48c3c47273c868c21169b3c91f0e0c], Pid [30048]"
````
In this example , the SYSLOG_HOSTNAME was set to "Rancher:" , and the container name is "rancher-agent".

[...]

Issue #137, opened 2015-11-03
MARATHON_APP_ID in log message


michallula:
Hi,
I wonder what is the best way to customize logs messages and to include additional information about Docker container from environment variables. I'm working on PaaS solution using Apache Mesos and Marathon. Marathon tasks launch Docker containers and set MARATHON_APP_ID variable which is human readable name of container. It can be retrieved with docker inspect command. I want to include this information in each log message. Is it possible with logspout?

scalp42:
+1 @progrium if you have any idea

orimarti:
Would be good being able to log some environment variables :)

progrium:
The templating you have has access to the entire Docker container structure
as far as I know. It should be possible.

On Thu, Feb 4, 2016 at 10:36 AM, Oriol Martí <notifications@github.com>
wrote:

> Would be good being able to log some environment variables :)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/gliderlabs/logspout/issues/137#issuecomment-179934465>
> .
>



-- 
Jeff Lindsay
http://progrium.com

vivekjuneja:
We are already using this in Production cluster. We use the Logstash adapter, and have [modified]([](https://github.com/vivekjuneja/logspout-logstash/blob/master/logstash.go)) it to use other useful environment variables set in the docker container.

PR #182, opened 2016-04-27
Add template function to syslog adapter to traverse .Container.Config.Env

Carles-Figuerola:
This was created as a way to pass a specific environment variable that's inside of the docker container. The variable .Container.Config.Env is formed of an array of key=value so it's not sortable or findable from the templating engine. Calling the container with something like this:
```
sudo docker run -d -v /var/run/docker.sock:/var/run/docker.sock -e SYSLOG_TAG="{{getEnvVar .Container.Config.Env \"MARATHON_APP_ID\"}}"  logspout:dev
```

will create a line as such:
```
Apr 27 15:50:56 87d5094fd8f8 nginx-marathon[19644]: 10.213.48.218 - - [27/Apr/2016:15:50:56 +0000] "GET / HTTP/1.1" 304 0 "-" "asdfsadf" "-"
```
where nginx-marathon is part of the .Container.Config.Env:
```
$ docker inspect 87d5094fd8f8 
<sic>
           "Env": [
                "MARATHON_APP_ID=nginx-marathon",
                "TEST=test",
                "SOMETHING_ELSE=false",
                "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
                "NGINX_VERSION=1.9.14-1~jessie"
            ],
<sic>
```

This would also help solve #163 and #137

PS. several envs can be also concatenated:
```
sudo docker run -d -v /var/run/docker.sock:/var/run/docker.sock -e SYSLOG_TAG="{{getEnvVar .Container.Config.Env \"MARATHON_APP_ID\"}}, {{getEnvVar .Container.Config.Env \"TEST\"}}"  logspout:dev
```

### DIFF
--- a/adapters/syslog/syslog.go
+++ b/adapters/syslog/syslog.go
@@ -8,6 +8,7 @@ import (
 	"log/syslog"
 	"net"
 	"os"
+	"strings"
 	"text/template"
 	"time"
 
@@ -27,6 +28,15 @@ func getopt(name, dfault string) string {
 		value = dfault
 	}
 	return value
+}
+
+func tplGetEnvVar(env []string, key string) string {
+        for _, value := range env {
+                if strings.HasPrefix(value, fmt.Sprintf("%s%s", key, "="))  {
+                        return strings.Split(value, "=")[1]
+                }
+        }
+        return ""
 }
 
 func NewSyslogAdapter(route *router.Route) (router.LogAdapter, error) {
@@ -67,7 +77,10 @@ func NewSyslogAdapter(route *router.Route) (router.LogAdapter, error) {
 	default:
 		return nil, errors.New("unsupported syslog format: " + format)
 	}
-	tmpl, err := template.New("syslog").Parse(tmplStr)
+
+	funcMap := template.FuncMap{
+		"getEnvVar":   tplGetEnvVar}
+	tmpl, err := template.New("syslog").Funcs(funcMap).Parse(tmplStr)
 	if err != nil {
 		return nil, err
 	}


adapters/syslog/syslog.go
```diff
@@ -29,6 +30,15 @@ func getopt(name, dfault string) string {
 	return value
 }
 
+func tplGetEnvVar(env []string, key string) string {
+        for _, value := range env {
+                if strings.HasPrefix(value, fmt.Sprintf("%s%s", key, "="))  {
```
koenbollen: Maybe do the `fmt.Sprintf("%s%s", key, "=")` once in the beginning of the function? Saves a few allocs.

adapters/syslog/syslog.go
```diff
@@ -67,7 +77,10 @@ func NewSyslogAdapter(route *router.Route) (router.LogAdapter, error) {
 	default:
 		return nil, errors.New("unsupported syslog format: " + format)
 	}
-	tmpl, err := template.New("syslog").Parse(tmplStr)
+
+	funcMap := template.FuncMap{
+		"getEnvVar":   tplGetEnvVar}
```
koenbollen: Would call the function `getenv` more in the line of other default template functions and the normal os.Getenv go function.

adapters/syslog/syslog.go
Carles-Figuerola: Thanks, changed.

adapters/syslog/syslog.go
Carles-Figuerola: I don't really understand your comment here. What I'm trying to do is create a new template function so you can get environment variables from the slice .Container.Config.Env , as you can see in the first code bit in my first comment.

koenbollen: Yeah, I understand. I just though it would be nice if the naming was more 'go-like'. Doesn't really matter much.
Cheers

JeanMertz:
@josegonzalez any chance to get this merged? This would help us a lot in identifying logs from specific services.

timbunce:
Ping. Another hopeful user.

adapters/syslog/syslog.go
```diff
@@ -29,6 +30,16 @@ func getopt(name, dfault string) string {
 	return value
 }
 
+func tplGetEnvVar(env []string, key string) string {
+        key_equals := fmt.Sprintf("%s%s", key, "=")
```
andrewgaul:
```go
key_equals := key + "="
```

adapters/syslog/syslog.go
```diff
@@ -29,6 +30,16 @@ func getopt(name, dfault string) string {
...
+        for _, value := range env {
+                if strings.HasPrefix(value, key_equals)  {
+                        return strings.Split(value, "=")[1]
```
andrewgaul:
```go
return value[strings.Index(value, "=")+1:]
```

adapters/syslog/syslog.go
```diff
@@ -29,6 +30,16 @@ func getopt(name, dfault string) string {
 	return value
 }
 
+func tplGetEnvVar(env []string, key string) string {
+        key_equals := key + "="
+        for _, value := range env {
+                if strings.HasPrefix(value, key_equals)  {
+                        return value[strings.Index(value, "=")+1:]
```
andrewgaul: Can simplify this to:

```go
return value[length(key_equals)+1:]
```

andrewgaul:
Could you merge this?  Otherwise I will need to create my own fork and push to Docker Hub.

billryan:
Hope this PR could be merged soon.

daniilyar:
+1, we really need this. Envvars are the only way to pass additional info about the container in case if your container orchestrator doesn't allow setting a container labels, but allows setting container envvars. The good example is Kubernetes, which does not allow (https://github.com/kubernetes/kubernetes/issues/3764) setting container labels, only pod labels

jorisw:
Did somebody try to merge this and the CI somehow failed? If this PR has been approved many of us would like to use it.

mashayev:
@jorisw
1) If I'm not wrong this feature was released. you can use:
    ```
    - name: SYSLOG_TAG
      value: '{{ index .Container.Config.Env "PROGRAM_NAME" }}'
    ```
    When `PROGRAM_NAME` is your environment variable.

2) In my environment the use case was different I just needed to know in which region I'm running, so I used configMap instead, I create configMap per environment and include the data for this specific environment.
    **For example:**
    Create configMap:
    ```
    ---
    apiVersion: v1
    kind: ConfigMap
    metadata:
      name: cluster-info
      namespace: kube-system
    data:
      name: amazon-web-services-us-east-1
    
    ```
    Use this configMap data in logspout yaml:
    ```
    env:
      - name: SYSLOG_HOSTNAME
        valueFrom:
          configMapKeyRef:
            name: cluster-info
            key: name
    ```

Hope this info was helpful for you :)

mmoore0011:
@mashayev 

When I try your solution I get 

 `error calling index: cannot index slice/array with type string`

This would be a great improvement for me.  I'm not sure if this commit dropped off the radar because no one is available to fix the test or because people have found other ways to include individual env vars in the log messages.  Does anyone following this have a working solution?

ksmithut:
@mmoore0011 Just ran into this from another issue here. I'm really close to having to do something like was suggested [here](https://github.com/gliderlabs/logspout/issues/202#issuecomment-232606393), but it's just essentially checking each character to see if it matches the environment variable name up until the `=`, the uses that. I've been working on one for the better part of an hour, but `Split` wasn't available...

Issue #187, opened 2016-05-12
Duplicate logs


patrick-bark:
Hey, have been using logspout for about a week now and am running into some pretty serious duplicate log issues. The simple script I am running is having its logs duplicated over 1000 times. I have verified that its not log stash or the application causing it. I have installed to logstash plugin for it, and I'm not sure if thats somehow messing it up. 

Here is my Dockerfile:

```
FROM gliderlabs/logspout:master
ENV SYSLOG_FORMAT rfc3164
```

and heres is my modules.go

```
package main

import (
    _ "github.com/gliderlabs/logspout/adapters/syslog"
    _ "github.com/gliderlabs/logspout/transports/tcp"
    _ "github.com/looplab/logspout-logstash"
    _ "github.com/gliderlabs/logspout/transports/udp"
)
```

I am using the latest image off of docker hub. Any idea what could be happening here.

josegonzalez:
Do you have a repro I can setup of the problem so we can investigate?

ebr:
Have been seeing the same for a while, on `master`.

It's very sporadic and next to impossible to reproduce. We occasionally get batches of 100,000+ identical log entries pushed to Loggly. I'm just using the syslog adapter. the loggly collector runs on the same host, listening on port 514. The `logspout:master` container is started with `CMD` `syslog://172.17.0.1:514`. I have not installed any extra plugins.

Stopping the logspout container and starting it again seems to help for a while (ranging from hours to days), but then it happens again and sometimes stops after a few minutes/hours, sometimes keeps pushing duplicate entries until I get an alert from Loggly about an overage.

I can provide any pertinent info about the environment, whatever you think may help.

ebr:
I think I'm able to narrow it down to cases when my app crashes inside the container, and the container is restarted (NOT recreated, but restarted). It probably makes sense as the same behaviour would be expected from `docker-compose logs -f`: all the logs of the container since its creation would be dumped to stdout.

josegonzalez:
Oh interesting. I'm not sure there is a good way around that if it's docker just spitting out the logs again.

ebr:
Thanks! I tried looking at the code myself, but it's way over my head so far :) Wondering if, when starting to tail the logs for a started container, it might be possible to not tail any preceding log entries (similar to `docker logs --tail=0`), and only follow the incoming entries.

josegonzalez:
[...]

PR #191, opened 2016-05-19
reconnect log stream unless container is dead

### LINKED ISSUE #168: Not sending logs to papertrail as of a few days ago
I've been running `latest` in my dev environments for a while now successfully and as of 3/3/2016 I noticed that we weren't getting any logs in papertrail.  The container is started and I when I do a `docker logs logspout` I see only the below (port changed for posting here).  I see nothing in docker.log or anywhere else.  I tried pulling/running the `v2` tag which works for a time and then ultimately crashes when another container is updated and possibly in other scenarios as well (haven't spent time troubleshooting `v2`...would rather get `v3` working).  Using `docker inspect logspout` I didn't see any obvious config differences between the running `v2` and the nonfunctional `v3`.  Any ideas or how I can further troubleshoot?

```
# logspout v3 by gliderlabs
# adapters: syslog raw tls udp tcp
# options : persist:/mnt/routes
# jobs    : pump http[logs,routes]:80
# routes  :
#   ADAPTER	ADDRESS				CONTAINERS	SOURCES	OPTIONS
#   syslog	logs2.papertrailapp.com:12345				map[]
```

bobzoller:
I believe this fixes #168 (at least my own root cause, which seems to be docker/docker#21220).

tested manually; will report back when I've tested it more widely.

### DIFF
--- a/router/pump.go
+++ b/router/pump.go
@@ -147,9 +147,15 @@ func (p *LogsPump) pumpLogs(event *docker.APIEvents, backlog bool) {
 	} else {
 		sinceTime = time.Now()
 	}
+
+	p.mu.Lock()
+	if _, exists := p.pumps[id]; exists {
+		p.mu.Unlock()
+		debug("pump.pumpLogs():", id, "pump exists")
+		return
+	}
 	outrd, outwr := io.Pipe()
 	errrd, errwr := io.Pipe()
-	p.mu.Lock()
 	p.pumps[id] = newContainerPump(container, outrd, errrd)
 	p.mu.Unlock()
 	p.update(event)
--- a/router/pump.go
+++ b/router/pump.go
@@ -140,11 +140,12 @@ func (p *LogsPump) pumpLogs(event *docker.APIEvents, backlog bool) {
 		debug("pump.pumpLogs():", id, "ignored: environ ignore")
 		return
 	}
-	var tail string
+
+	var sinceTime time.Time
 	if backlog {
-		tail = "all"
+		sinceTime = time.Unix(0, 0)
 	} else {
-		tail = "0"
+		sinceTime = time.Now()
 	}
 	outrd, outwr := io.Pipe()
 	errrd, errwr := io.Pipe()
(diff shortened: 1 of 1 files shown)


johanndt:
Hi @bobzoller - any more results from your testing? Do you have an image up on docker hub I can also try?

johanndt:
For info, I searched around a bit and found an image at https://hub.docker.com/r/goodeggs/logspout/tags/ with the `reconnect-logs` tag. We've been running it and while some containers seem to be faring better, we are still losing some in papertrail after about half a day.

router/pump.go
```diff
@@ -140,37 +140,58 @@ func (p *LogsPump) pumpLogs(event *docker.APIEvents, backlog bool) {
...
+
+			sinceTime = time.Now()
+
+			container, err := p.client.InspectContainer(id)
+			if err != nil {
+				_, four04 := err.(*docker.NoSuchContainer)
+				if !four04 {
+					assert(err, "pump")
+				}
+			} else if container.State.Running {
```
rdbaron: I found a problem here when testing this fix.  My container restarts so quickly that this check always returns true.  This causes the deletion of the pump below to be skipped and each time my container restarts I get a new one.  This leads to LOTS of duplication.

bobzoller: I haven't run into this problem, but it makes perfect sense.  Off the cuff it seems like we could add a check around line 153 and only create a new pump if one does not exist for that container.  eg `if _, ok := p.pumps[id]; !ok { }`.  Happy to be corrected.

josegonzalez: That should work, mind modifying and testing out that solution?

bobzoller:
I've pushed a new version of docker image `goodeggs/logspout:reconnect-logs` with the fix in 13d1129.  It's running fine on my dev cluster, but perhaps @rdbaron could attempt to reproduce the issue and let us know if it is fixed?

rdbaron:
@bobzoller: will do, I'll let you know once I have.

rdbaron:
@bobzoller; I can confirm that this fixes the issue with duplicating the pump when the container restarts quickly.

bobzoller:
great, thanks!  @josegonzalez seems like this might be ready to merge.

josegonzalez:
Works for me :)

josegonzalez:
Thanks to everyone for being diligent about this PR. Please direct me to any other ones we need to shepherd to greatness :)

mattaitchison:
I am working on a solution for #187 which may also take care of this. 

I also think there are some edge cases this doesn't take care of.

josegonzalez:
@mattaitchison Feel free to override me on this PR, I consider logspout your baby :)

bobzoller:
yes @mattaitchison , and I was fairly certain there were some edge cases this didn't take care of ;)

PR #198, opened 2016-06-02
Much, much faster builds.

treeder:
This fixes the really slow build process on both building logspout directly as well as when a user is adding a module (ONBUILD). 

Part of it is vendoring, part of it is removing a bunch of build steps (in build.sh) that only need to happen once, not every build.


progrium:
I'm down with vendoring except I'd like to figure out a way to test both building with latest and with vendor'd deps. Haven't standardized on a vendoring/dep tool. If you're going to sneak in glide can we have a (hopefully short) discussion on glider vs godep?

treeder:
Yes, definitely, all these pull requests are totally open for discussion. Regarding glide vs godep , glide is much better, just give it a try and you'll see. Also , doesn't require committing the vendor directory, just the two glide files. So it's sort of optional.

josegonzalez:
@treeder mind taking a look at the test fails?

treeder:
I'll check this out soon.

treeder:
Fixed circle and rebased.

michaelshobbs:
I'm 👍  on `glide` vs `godep`. There's also `dep` that the core team is working on but I don't have any idea when that will see the light of day. Not sure it's worth waiting for at this point either as we can also move to it when the time is right. 

What's more interesting to me about this PR is the change in base image. Is everybody onboard with that? If so, I'd like to see this PR rebased and merged sooner rather than later and I'm happy to help with that effort.

treeder:
The core dep tool is almost the same as glide btw, same concepts. 

And regarding base image, this PR should probably switch to the official images now that they all have alpine variants, eg: `golang:1.8-alpine`

michaelshobbs:
@treeder can you update this PR or shall I run with it?

treeder:
@michaelshobbs if you want to fix this up, fine by me. Sorry for delay.

michaelshobbs:
@treeder Perhaps my noobness is showing but I don't see where this actually uses glide to retrieve the dependencies. Am I missing something?

PR #200, opened 2016-06-06
Make Build Script Extentable

digitalkaoz:
since we use self signed certificates here, we have no option to install those certificates before the `build-sh` step. so we make it possible to provide an own `build.sh` script and run that instead of the vendored one...


josegonzalez:
Does this break if that file doesnt exist?

josegonzalez:
*bump*

digitalkaoz:
sorry for the delay, no it doesnt break if the file isnt there...

PR #201, opened 2016-06-13
Fix duplicate logs & backlogs on container restart

ebr:
Connects to #187 
- Nothing needs to be done on "restart" event, because a "start" event always(?) follows it. A `pumpLogs` was being called on both `restart` and `start`, which caused duplicate entries after a restart.
- Add support for a `BACKLOGS` environment variable:
  - Unless specified `BACKLOGS=false`, it will default to `true` for all other values, consistent with current behaviour;
  - `BACKLOGS=false` is equal to `docker logs --tail=0`


olivergrahl:
+1 for getting this merged. We start containers using cron every couple of minutes, so we get all the log history sent again on every single start.

olivergrahl:
Any news on this? It's really an issue for us, as we start containers every couple of minutes, and monitor the logs via Papertrail alerts. Once a string that would trigger an alert gets logged from the container, it will be logged again and again on each container start, resulting in the alert being triggered again and again... Any idea how we can prevent that? Thx!

ebr:
@josegonzalez are we still looking at #209? Shall I rebase this PR, or abandon it?

ebr:
@josegonzalez (cc @olivergrahl) this is rebased, tested in my Staging and Prod and looking good. Takes care of the backlog dump with `BACKLOG=false`. double pump was already fixed elsewhere.

michaelshobbs:
merged in #280

PR #203, opened 2016-07-01
ignore containers with unsupported log drivers

pmbauer:
trying to hit /logs for a container that does not support that endpoint
results in the following error in that container's stdout

"Error running logs job: configured logging reader does not support
reading"

At least in our production environment (docker 1.9.1) the problem is amplified.  Logspout attempts to connect to the container's /log endpoint repeatedly resulting in a flood of such messages in the stdout of containers that are configured with the syslog driver.

This commit fixes that behavior and makes the supported log drivers explicit.


josegonzalez:
Thanks for the pr!

PR #204, opened 2016-07-05
Use InactivityTimeout to work around a bug in docker

markine:
Log rotation causes 'docker logs --follow' to stop outputting logs (docker/docker#23913).

We work around this by detecting inactivity in a log stream and restarting its pump. We restart the pump with argument Since=Now-InactivityTimeout to catch any logs we may have "missed" due to log rotation (this is harmless for truly silent log streams). We may still lose some logs from an app that's logging a lot due to small delays in recognizing an inactive stream on the Docker side and restarting a pump on the Logspout side.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,10 @@ Routes are stored on disk, so by default routes are ephemeral. You can mount a v
 
 See [routesapi module](http://github.com/gliderlabs/logspout/blob/master/routesapi) for all options.
 
+#### Detecting timeouts in Docker log streams
+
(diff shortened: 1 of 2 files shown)


josegonzalez:
Can you add some documentation for this?

markine:
@josegonzalez certainly, what do you have in mind? Where is a good place to add documentation about the new env var?

josegonzalez:
@markine Probably a new section under [here](https://github.com/gliderlabs/logspout#using-logspout).

markine:
@josegonzalez Done, please take a look.

josegonzalez:
@mattaitchison this looks like it will collide with your PR, though I think it's definitely worth pulling in because docker lol. Thoughts on merging this first? I can def rebase your PR if that makes it simpler for you.

rosskukulinski:
This has resolved the missing log files for my client.

markine:
Ping?

markine:
ping @josegonzalez

paulanunda:
Is this going to be merged soon? It would be quite useful for those running on Kubernetes.

timbunce:
Any reason this isn't enabled by default?

PR #206, opened 2016-07-06
Add custom CA and client certificates basic support for tls transport

bn0ir:
Add custom ca basic tls transport support:
- default ca path: /mnt/ca/;
- ca path can be set via CA_PATH env;

Add basic client certificates tls transport support:
- default certificates/keys path: /mnt/cert/;
- certificates/keys path can be set via CERT_PATH env;
- certificate names: .crt, .cert; key names: .key;
- example: user.cert, user.key

PR #209, opened 2016-07-11
WIP: Now using event timestamp to retrieve relevant logs.

mattaitchison:
Using the event timestamp should prevent duplicate logs when a container restarts.

### DIFF
--- a/router/pump.go
+++ b/router/pump.go
@@ -120,7 +120,7 @@ func (p *LogsPump) Run() error {
 		p.pumpLogs(&docker.APIEvents{
 			ID:     normalID(listing.ID),
 			Status: "start",
-		}, false)
+		})
 	}
 	events := make(chan *docker.APIEvents)
 	err = p.client.AddEventListener(events)
@@ -131,7 +131,7 @@ func (p *LogsPump) Run() error {
 		debug("pump.Run() event:", normalID(event.ID), event.Status)
 		switch event.Status {
 		case "start", "restart":
-			go p.pumpLogs(event, true)
+			go p.pumpLogs(event)
 		case "rename":
 			go p.rename(event)
 		case "die":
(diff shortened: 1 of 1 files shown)


router/pump.go
```diff
@@ -153,67 +153,25 @@ func (p *LogsPump) pumpLogs(event *docker.APIEvents, backlog bool) {
 		debug("pump.pumpLogs():", id, "ignored: environ ignore")
 		return
 	}
-	if !logDriverSupported(container) {
```
josegonzalez: This drops the following functionality https://github.com/gliderlabs/logspout/commit/324db6e01dda3aa6727c03cb3f29217728bd861f.

mattaitchison: Damn it. I didn't mean to remove that commit.

router/pump.go
```diff
@@ -153,67 +153,25 @@ func (p *LogsPump) pumpLogs(event *docker.APIEvents, backlog bool) {
...
-		debug("pump.pumpLogs():", id, "pump exists")
-		return
-	}
 	outrd, outwr := io.Pipe()
 	errrd, errwr := io.Pipe()
 	p.pumps[id] = newContainerPump(container, outrd, errrd)
 	p.mu.Unlock()
 	p.update(event)
 	go func() {
-		for {
```
josegonzalez: This kills the work done to fix the fact that we previously didn't reconnect, as mentioned here: https://github.com/gliderlabs/logspout/pull/191

ebr:
How is it looking for getting this merged (without breaking the work done in #191)? 
I'm between a rock and a hard place: would like to run `master` in hopes for a fix for #206, but I need to run my fork based on #201 in order to avoid duplication.

michaelshobbs:
@mattaitchison do you still want to pursue this route?

mattaitchison:
@michaelshobbs yeah I do. I believe its a better route since it should prevent missing log output as described [here](https://github.com/gliderlabs/logspout/commit/3b7bae6ebab8360068533bbbd7265e3a5a6a149c#diff-04c6e90faac2675aa89e2176d2eec7d8R101).

PR #222, opened 2016-08-10
added the ability to include only container having label using INCLUDE_LABEL param

mohamed-el-habib:
Added the ability to include only container having label using INCLUDE_LABEL param.

### DIFF
--- a/router/pump.go
+++ b/router/pump.go
@@ -73,6 +73,16 @@ func ignoreContainer(container *docker.Container) bool {
 	if value, ok := container.Config.Labels[excludeLabel]; ok {
 		return len(excludeLabel) > 0 && strings.ToLower(value) == "true"
 	}
+	includeLabel := getopt("INCLUDE_LABEL", "")
+	if includeLabel != "" {
+		if _, ok := container.Config.Labels[includeLabel]; ok {
+			// include all containers that have the includeLabel
+			return false
+		} else {
+			// ignore all containers that don't have the includeLabel
+			return true
+		}
+	}
 	return false
 }
 


josegonzalez:
Don't we have this somewhere already?

mohamed-el-habib:
Maybe i missed that, but as i saw there are some things similar when logs are routed, but no things when logs are pumped.

router/pump.go
```diff
@@ -73,6 +73,16 @@ func ignoreContainer(container *docker.Container) bool {
 	if value, ok := container.Config.Labels[excludeLabel]; ok {
 		return len(excludeLabel) > 0 && strings.ToLower(value) == "true"
 	}
+	includeLabel := getopt("INCLUDE_LABEL", "")
+	if includeLabel != "" {
+		if _, ok := container.Config.Labels[includeLabel]; ok {
+			// include all containers that have the includeLabel
```
josegonzalez: Comments shouldn't be necessary, and if they are, the code needs to be made more clear.

router/pump.go
```diff
@@ -73,6 +73,16 @@ func ignoreContainer(container *docker.Container) bool {
...
+			return false
+		} else {
```
josegonzalez: We don't need the explicit else if the above will return false.

router/pump.go
```diff
@@ -73,6 +73,16 @@ func ignoreContainer(container *docker.Container) bool {
 	if value, ok := container.Config.Labels[excludeLabel]; ok {
 		return len(excludeLabel) > 0 && strings.ToLower(value) == "true"
 	}
+	includeLabel := getopt("INCLUDE_LABEL", "")
+	if includeLabel != "" {
+		if _, ok := container.Config.Labels[includeLabel]; ok {
```
josegonzalez: Is there a reason this check isn't similar to the exclude check where we check the actual value?

mohamed-el-habib:
@josegonzalez : I simplified the code as i can (i my first time with golang), and updated the README.md.

mohamed-el-habib:
@josegonzalez did my code need documentation ?

itajaja:
what's the difference between this vs the filter approach?

mohamed-el-habib:
May be, misunderstood the filter approach. My use case is to have a docker-compose file with a list of containers for my application. And I want to add logspout container to this compose file. I want logspout to load only the compose container 's logs and expose them throw http module. How can I do that using filter approach.

michaelshobbs:
@mattaitchison @josegonzalez Thoughts on having both `INCLUDE` & `EXCLUDE`?

PR #225, opened 2016-08-19
FIX : add build-base package install to fix missing gcc

ekkinox:
FIX : add build-base package install to fix missing gcc

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,13 +1,13 @@
 #!/bin/sh
 set -e
-apk add --update go git mercurial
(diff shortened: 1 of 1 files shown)


pboyer:
Really looking forward to this one. :)

pboyer:
I hate to be peppering this PR with "Me toos", but I'm pretty sure this fix is required to do any adapter development for logspout - a defining feature of the tool! It would be great to see it merged. :)

Issue #210, opened 2016-07-13
Log pump closes after single send timeout


drewrobb:
In prod we have noticed a single send timeout causes logs to stop being sent. In logspout logs we see messages like `2016/07/12 07:44:54 pump.send(): send timeout, closing` and then no more logs from that container are sent. 

The relevant code is here: https://github.com/markine/logspout/blob/324db6e01dda3aa6727c03cb3f29217728bd861f/router/pump.go#L332

Is it intended for the logstream to be restarted by some other mechanism? Looking to where this change was introduced, there was previously no `delete` here and a `// TODO: log err after timeout and continue`-- https://github.com/markine/logspout/commit/1d54f5e984f30bb38b7a9754333c15e80fd780da#diff-1e402098f7c31429ceb791bb7bbb6620L216

It seems to me that the behavior of deleting a logstream from a pump after a single send timeout is unreasonable if there is no other mechanism to restart it?

josegonzalez:
What version of logspout are you running?

drewrobb:
Logspout master branch. We are also using a fork of  https://github.com/rtoma/logspout-redis-logstash as an adaptor, and that is the source of timeouts. It appears that this commit: https://github.com/rtoma/logspout-redis-logstash/commit/b96fbb63fa275395647ce01601f80d645227e591#diff-bbc64daf788a83d9a6daf7056d49d2a8R115 actually addresses this issue by adding timeouts to satisfy the 1 second timeout in logspout. However we do not have not have this commit included in the version of logspout-redis-logstash we are running. Having this commit in our production version would probably fix our issue, however I think the underlying behavior of logspout is not good.

ebr:
We are seeing similar/same behaviour, suspecting this started after upgrading docker to v1.12. I also suspect #204 is a contributing factor.

ebr:
@drewrobb are you still sure the send timeout is the cause of this? I've narrowed it down to log rotation; we do get send timeouts, but i can't directly correlate them with this.

drewrobb:
@ebr we do not have docker log rotation setup and still (had previously) send timeouts. We updated to the latest logspout-redis-logstash and the comment I mentioned in the issue description didn't solve our problem. We are running this branch now: https://github.com/gliderlabs/logspout/compare/master...strava:dr/no-delete-logstream which finally fixed it for us. We are also on docker 1.12.

dareid:
I think we are seeing similar behaviour to this, any news on a fix being merged?

[...]

PR #235, opened 2016-09-23
defer delete logstream

josegonzalez:
Don't know why it was commented out. Refs #210


dareid:
Any news on this pull request?

fjollberg:
Funny, just saw this. Have been working around issues with sending logs for weeks, thought it was the log rotation issue, but turned out that it wasn't only this.

I'm using logspout to forward messages to Azure OMS and quite often hit this timeout, causing my logs to stop propagating. My workaround was to raise the timeout to 60s in my own fork since I wasn't quite sure it was safe to remove the deletion entirely. Anyhow, I figure this would solve my remaining issues.

michaelshobbs:
Going to pull in another PR that deals with this

Issue #241, opened 2016-11-03
ERROR: the correct import path is gopkg.in/check.v1 


jalberto:
cannot build becuase of incorrect import path

amitripshtos:
I can not build anymore too :(

KoenR3:
Same problem here

ccadruvi:
Switching to gliderlabs/alpine:latest in the Dockerfile fixed the build issue for me.

david-kli:
I encountered this error when trying to build a custom logspout image using a new module.  To rule out any issues with the new module, I cloned the logspout repo ran a docker build in the custom directory.  I was able to get the same error:

```
# github.com/go-check/check
../../go-check/check/error.go:4: "ERROR: the correct import path is gopkg.in/check.v1 ... " evaluated but not used
```

Any help would be greatly appreciated

Here is the full output:
**cd logspout/custom**
**docker build -t test1 .**

```
Sending build context to Docker daemon 5.632 kB
Step 1 : FROM gliderlabs/logspout:master
# Executing 3 build triggers...
Step 1 : COPY ./build.sh /src/build.sh
 ---> Using cache
Step 1 : COPY ./modules.go /src/modules.go
Step 1 : RUN cd /src && ./build.sh "$(cat VERSION)-custom"
 ---> Running in d5c5d56d0d50
fetch http://alpine.gliderlabs.com/alpine/v3.3/main/x86_64/APKINDEX.tar.gz
fetch http://alpine.gliderlabs.com/alpine/v3.3/community/x86_64/APKINDEX.tar.gz
(1/39) Upgrading musl (1.1.12-r6 -> 1.1.12-r7)
(2/39) Installing binutils-libs (2.25.1-r0)
(3/39) Installing binutils (2.25.1-r0)
(4/39) Installing gmp (6.1.0-r0)
(5/39) Installing isl (0.14.1-r0)
(6/39) Installing libgomp (5.3.0-r0)
(7/39) Installing libatomic (5.3.0-r0)
(8/39) Installing libgcc (5.3.0-r0)
(9/39) Installing pkgconf (0.9.12-r0)
(10/39) Installing pkgconfig (0.25-r1)
(11/39) Installing mpfr3 (3.1.2-r0)
(12/39) Installing mpc1 (1.0.3-r0)
(13/39) Installing libstdc++ (5.3.0-r0)
(14/39) Installing gcc (5.3.0-r0)
(15/39) Installing make (4.1-r0)
[...]

PR #242, opened 2016-11-08
Updating Dockerfile base image to gliderlabs/alpine:3.4 from 3.3

david-kli:
This was done to upgrade the go version and resolve a go build import
error for go-check

Fix for logspout issue #241


JensPiegsa:
Is there going to be a new docker hub build reflecting this change?

Issue #234, opened 2016-09-19
Duplicate Logs After Docker Image Restart


mbrooks:
When a container is running and is restarted, all of the stdout for the docker container that is being restarted is sent back through logspout and logged again. 

Has anyone else seen this behavior? Let me know if I need to collect any debugging, or give any detail towards the issue. I could possibly also create a patch if pointed in the right direction.

kamrar:
Same problem here with logspout v3.2-dev.

iravench:
Observed the same behaviour here

ebr:
@mbrooks, see #187 as well as a connected PR #201, which provides a fix. unfortunately, i do not know if it will ever be merged.

erikverheij:
Same problem here. As a workaround we use the `--force-recreate` option with `docker-compose`. On container recreation the logs are cleared. Unfortunately this will not work on automatic restarts.

There is also a [feature request in docker-compose](https://github.com/docker/compose/issues/1083) to clear container logs.

michaelshobbs:
Closing as this is covered in a couple PRs

PR #252, opened 2017-01-06
Disable backlog on container start/restart.

roman-vynar:
Addresses https://github.com/gliderlabs/logspout/issues/187, https://github.com/gliderlabs/logspout/issues/234

Sending the backlog does not really make sense on container restart or stop/start. Usually, you have those logs already and do not need duplicate ones.

### DIFF
--- a/router/pump.go
+++ b/router/pump.go
@@ -140,7 +140,7 @@ func (p *LogsPump) Run() error {
 		debug("pump.Run() event:", normalID(event.ID), event.Status)
 		switch event.Status {
 		case "start", "restart":
-			go p.pumpLogs(event, true, inactivityTimeout)
+			go p.pumpLogs(event, false, inactivityTimeout)
 		case "rename":
 			go p.rename(event)
 		case "die":


ebr:
@roman-vynar: it was previously mentioned in #187 and/or #201 that backlogs are needed on start because the container might have output logs before logspout has had a chance to connect. That's why #201 implements this as an env. var flag.
(off topic: your city is the loveliest place in the world :))

roman-vynar:
@ebr then 2 things:
- what if there are too much pending logs at that moment, sending them all could very expensive
- when syslog connection is failed, we do not send logs anyway and obviously skip them

So the flag does not solve the above 2 issues. I would say it is better to have an option to send existing logs somehow else if needed on start.

P.S. offtopic: hah, have you been there? 😆

ebr:
@roman-vynar i agree this PR fixes the issue, though in my opinion it's better to provide a runtime flag, in case one relies on being certain that no log lines are lost from the beginning of the container start (and is prepared to deal with expensive sending & duplicates). But this is up to maintainers to decide, of course.

offtop: i grew up there :)

router/pump.go
```diff
@@ -140,7 +140,7 @@ func (p *LogsPump) Run() error {
 		debug("pump.Run() event:", normalID(event.ID), event.Status)
 		switch event.Status {
 		case "start", "restart":
-			go p.pumpLogs(event, true, inactivityTimeout)
+			go p.pumpLogs(event, false, inactivityTimeout)
```
michaelshobbs: Perhaps we should only set this to `false` on restart and have a `true` case on start. Thoughts?

michaelshobbs: Just saw your previous conversation. I think I'd like to keep backlog set to `true` on start for now. That at least then doesn't break the current expectation.

roman-vynar: We can create an option then, if someone needs to re-send them.

Leaving it on `start` will not help as you can stop/start and it is all sent again.

michaelshobbs: An option is fine with me.

michaelshobbs: @roman-vynar is that something you're able to add here? if not, no worries but wanted to check before I run with this.

davidnortonjr:
@michaelshobbs  #201 works great in my fork (just because somebody already did the work to make it configurable). My vote would be to merge that one.

michaelshobbs:
I concur. I'll go with #201 unless @roman-vynar has any objection

roman-vynar:
@michaelshobbs thanks 👍

PR #261, opened 2017-02-11
Adapter for GELF (Graylog Extended Log Format)

mikaelcabot:
GELF: http://docs.graylog.org/en/2.1/pages/gelf.html

Tested with Graylog version: 2.1.3-1

(Made with Inspiration from the syslog adapter)


michaelshobbs:
The current thinking is to not include implementation specific adapters in the core repo. You can follow the pattern like [logspout-logstash](https://github.com/looplab/logspout-logstash) and folks can follow the instructions [here](https://github.com/gliderlabs/logspout/blob/26966592a2872c61c741934c91759d1da90ab545/MODULES.md) for building and running custom adapters.

If we get more maintainers onboard, I'm open to discussing the inclusion of specific adapters.

PR #265, opened 2017-02-14
Context error

joaquincasares:
I was running into this error when attempting to use the [logspout-logstash](https://github.com/looplab/logspout-logstash) plugin:

```
+ go get
package context: unrecognized import path "context" (import path does not begin with hostname)
ERROR: Service 'elk_logspout' failed to build: The command '/bin/sh -c cd /src && ./build.sh "$(cat VERSION)-custom"' returned a non-zero code: 1
```

After searching for a solution, I found that the package used to be named `net/context` in Go 1.6 and simply `context` in Go 1.7.

I'm unsure if I made the most ideal changes, but I've included my personal workaround until there's an official fix.

Thanks for the awesome software!

### DIFF
--- a/custom/build.sh
+++ b/custom/build.sh
@@ -1,6 +1,8 @@
 #!/bin/sh
 set -e
-apk add --update go git mercurial build-base ca-certificates
+echo "http://alpine.gliderlabs.com/alpine/v3.5/community" \
+    >> /etc/apk/repositories
+apk add --update 'go>1.7' git mercurial build-base ca-certificates
 mkdir -p /go/src/github.com/gliderlabs
 cp -r /src /go/src/github.com/gliderlabs/logspout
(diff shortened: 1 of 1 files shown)


mattaitchison:
Thanks for the pr! 

Are you able to see if changing base image to `alpine:3.5` in the [Dockerfile](https://github.com/gliderlabs/logspout/blob/master/Dockerfile#L1) will fix this issue? You may need to modify the base image used in `custom/Dockerfile` to test it.

joaquincasares:
Simply replacing the base image didn't work since there's no logic to run `build.sh`. However, I did confirm that `apk add --update 'go>1.7'` works as expected.

michaelshobbs:
I checked locally and the move to `alpine:3.5` is *now* installing `go 1.7.3`.

PR #269, opened 2017-02-22
Add Rancher stack and service to structured data

mccricardo:
Implements [Add stack and service to logspout](https://jira.uphold.com/browse/INFRA-1167).

Also includes change that does not pump full log on a restart.


mccricardo:
Unwanted PR.

PR #276, opened 2017-03-27
add linting and fix lintballs

michaelshobbs:
lint with `golint` and `go tool vet`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,15 +22,12 @@ build:
 	docker build -t $(NAME):$(VERSION) .
 	docker save $(NAME):$(VERSION) | gzip -9 > build/$(NAME)_$(VERSION).tgz
 
-lint-direct:
-	go install \
+lint:
+	test -x $(GOPATH)/bin/golint || go get github.com/golang/lint/golint
+	go get \
+		&& go install \
 		&& ls -d */ | egrep -v '/custom/' | xargs $(XARGS_ARG) go tool vet -v
 	@if [ -n "$(shell $(GOLINT))" ]; then $(GOLINT) && exit 1; fi
-
-lint: build-dev
-	docker run \
-		-v $(PWD):/go/src/github.com/gliderlabs/logspout \
-		$(NAME):dev make -e lint-direct
 
 test: build-dev
 	docker run \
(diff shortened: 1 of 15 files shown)


adapters/syslog/syslog.go
```diff
@@ -112,7 +115,7 @@ func (a *SyslogAdapter) Stream(logstream chan *router.Message) {
 	}
 }
 
-func (a *SyslogAdapter) retry(buf []byte, err error) error {
+func (a *Adapter) retry(buf []byte, err error) error {
```
josegonzalez: Should internal functions also have a docblock?

michaelshobbs: It's not required or enforceable. However, if a function is doing something complex or "needs" additional explanation, then I tend to add them in that case.

progrium:
Can you simplify the Makefile? Maybe less Bash scripting?

michaelshobbs:
@progrium i played with this quite a bit. open to suggestions.

Makefile
```diff
@@ -18,6 +22,13 @@ build:
 	docker build -t $(NAME):$(VERSION) .
 	docker save $(NAME):$(VERSION) | gzip -9 > build/$(NAME)_$(VERSION).tgz
 
+lint:
+	test -x $(GOPATH)/bin/golint || go get github.com/golang/lint/golint
+	go get \
+		&& go install \
+		&& ls -d */ | egrep -v '/custom/' | xargs $(XARGS_ARG) go tool vet -v
+	@if [ -n "$(shell $(GOLINT))" ]; then $(GOLINT) && exit 1; fi
```
josegonzalez: `ifneq ($(GOLINT),)` will do.

michaelshobbs: if blocks run on all targets. i don't think we want that.

josegonzalez: Yeah but the var is already always set, no?

josegonzalez: Oh nvm `:=`. Dumb me.

adapters/syslog/syslog.go
josegonzalez: cool cool

Makefile
michaelshobbs: Yeah `GOLINT` is a string that only gets processed in the `lint` target

josegonzalez:
We can simplify the makefile at a later date. Getting linting working is probably the first priority :)

josegonzalez:
By we I mean @michaelshobbs (lord knows I'm not touching this).

PR #278, opened 2017-03-30
set exec bit on build.sh

### LINKED ISSUE #238: ./build.sh: Permission denied
Environment: 
- Windows 10
- Docker version 1.12.1
- docker-compose version 1.8.0 and Docker Hub

I had the [missing gcc error](https://github.com/gliderlabs/logspout/issues/223) and did the workaround of using gliderlabs/logspout:master and putting a copy of build.sh in my local directory. Now when I try to build, I get an error of `./build.sh: Permission denied`. You can see full build log at https://hub.docker.com/r/nikki603/logspoutkinesis/builds/bchksm9an3agrmdxavr5o78/

I get the same error when running docker-compose locally.

```
docker-compose build logspout
Building logspout
Step 1 : FROM gliderlabs/logspout:master
# Executing 3 build triggers...
Step 1 : COPY ./build.sh /src/build.sh
Step 1 : COPY ./modules.go /src/modules.go
Step 1 : RUN cd /src && ./build.sh "$(cat VERSION)-custom"
 ---> Running in 48852ae2f7b1
/bin/sh: ./build.sh: Permission denied
ERROR: Service 'logspout' failed to build: The command '/bin/sh -c cd /src && ./build.sh "$(cat VERSION)-custom"' returned a non-zero code: 126
```

build.sh file permission locally are -rwxr-xr-x

I believe this is related to bug https://github.com/docker/compose/issues/3716

michaelshobbs:
fixes #238

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,4 +8,4 @@ RUN cd /src && ./build.sh "$(cat VERSION)"
 
 ONBUILD COPY ./build.sh /src/build.sh
 ONBUILD COPY ./modules.go /src/modules.go
-ONBUILD RUN cd /src && ./build.sh "$(cat VERSION)-custom"
+ONBUILD RUN cd /src && chmod +x ./build.sh && ./build.sh "$(cat VERSION)-custom"


Chris7:
This fix introduces this bug:

https://github.com/docker/docker/issues/9547

The solution is to run `sync` after `chmod`

michaelshobbs:
@Chris7 mind putting a PR together to implement the fix?

Chris7:
@michaelshobbs https://github.com/gliderlabs/logspout/commit/10088518dbfc6f84073126e836cb0a9afe31d762#diff-3254677a7917c6c01f55212f86c57fbf actually fixed this bug, but that isn't what was on the `:master` tag on dockerhub so I encountered it. Building an image from the latest master myself and using my own registry circumvents this issue for now.

michaelshobbs:
Yeah I forgot about this one after I merged the build refactor. I'm looking into the docker hub build issues now.

haggaret:
Was this ever fixed? We're still seeing issues trying to build on docker hub when referencing the master version of logspout.

michaelshobbs:
The file is checked in with `+x` in master and builds fine in our CI.  If you're still seeing an issue, please open a new issue and include as much detail as possible; including any terminal output and errors.

haggaret:
Thanks @michaelshobbs, appreciate the quick response. Noticed that the change in this PR was to do the chmod inside the Dockerfile, but from the sounds of things (and looking at the latest master), this was changed to simply check in the build.sh file as executable and then it wasn't necessary to do the chmod in the Dockerfile.

It seems that when we reference your image in our Dockerfile (FROM gliderlabs/logspout:master), we need to include your build.sh file (and make it executable), which seems strange to me, but perhaps that expected...

PR #287, opened 2017-04-11
Consider ECONNRESET a non-temporary error.

luketurner:
This PR ensures that the `syslog` adapter treats ECONNRESET errors as "non-temporary", meaning that we immediately recreate the connection instead of first attempting to re-send the failed packets on the existing connection.

It should be noted, there are cases ("under certain rapid connection setup/teardown conditions") where ECONNRESET *can* be a temporary error. See the [Golang issue](https://github.com/golang/go/issues/6163) and associated [commit](https://github.com/golang/go/commit/a07a57b00ec9fdd8f6b02360d39454859709d08a). 

However, it seems to me that of the real-world cases where an ECONNRESET is encountered, the great majority of the problems would be non-temporary. For example, Papertrail delivers an ECONNRESET after 15 minutes of inactivity, which indicates that they have permanently closed the connection and we need to open a new one.

### DIFF
--- a/adapters/syslog/syslog.go
+++ b/adapters/syslog/syslog.go
@@ -9,6 +9,7 @@ import (
 	"net"
 	"os"
 	"strconv"
+	"syscall"
 	"text/template"
 	"time"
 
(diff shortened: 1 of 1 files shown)


mattaitchison:
I think we should try using write/read deadlines before accepting this.  Ive included a link to docs about SetDeadline.


```
        // SetDeadline sets the read and write deadlines associated
        // with the connection. It is equivalent to calling both
        // SetReadDeadline and SetWriteDeadline.
        //
        // A deadline is an absolute time after which I/O operations
        // fail with a timeout (see type Error) instead of
        // blocking. The deadline applies to all future and pending
        // I/O, not just the immediately following call to Read or
        // Write. After a deadline has been exceeded, the connection
        // can be refreshed by setting a deadline in the future.
        //
        // An idle timeout can be implemented by repeatedly extending
        // the deadline after successful Read or Write calls.
        //
        // A zero value for t means I/O operations will not time out.
        SetDeadline(t time.Time) error
```
https://golang.org/pkg/net/#Conn

Issue #277, opened 2017-03-30
logspout stops sending logs via tcp


akosyrev:
After 10-20mins after start container does not send any packets.

My problem was sending containers logs to logstash.
1.  I did changes to modules.go as described [here ](https://github.com/looplab/logspout-logstash/blob/master/README.md
)
2. Compile logspout on my host: `GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build` 
```
go version
go version go1.7.4 linux/amd64
```
3. Add binary to /bin

After start container sends logs well, no drops or anything wrong. But after sometime it stops sending packets, I see it in CPU and network usage. My main container (which generates logs to stdout) works, its responses for requests, writes logs..

I verified if it is logstash-module problem. I ran luisbebop/echo-serverPorts and tried raw TCP sending and faced same beavior.

Any help is appreciated

**UPD1**
Restart logspout-container fixes problem for 10-20 mins.

This is my logsout output after start:
```
30.03.2017 13:11:56# logspout  by gliderlabs
30.03.2017 13:11:56# adapters: syslog tcp udp tls logstash raw
30.03.2017 13:11:56# options : persist:/mnt/routes
30.03.2017 13:11:56# jobs    : routes http[logs,routes]:80 pump
30.03.2017 13:11:56# routes  :
30.03.2017 13:11:56#   ADAPTER	ADDRESS		CONTAINERS	SOURCES	OPTIONS
30.03.2017 13:11:56#   raw+tcp	10.6.109.62:8800*load*			map[]
```

michaelshobbs:
What version are you running? Can you also post the logspout container logs after you stop seeing packets being sent?

michaelshobbs:
potentially related to #268

akosyrev:
Version: v3.2-dev

There is no other logs after packets stop being sent.
Only this lines after startup:
```
31.03.2017 9:51:42# logspout  by gliderlabs
31.03.2017 9:51:42# adapters: syslog tcp udp tls logstash raw
31.03.2017 9:51:42# options : persist:/mnt/routes
31.03.2017 9:51:42# jobs    : http[routes,logs]:80 pump routes
31.03.2017 9:51:42# routes  :
31.03.2017 9:51:42#   ADAPTER	ADDRESS		CONTAINERS	SOURCES	OPTIONS
31.03.2017 9:51:42#   raw+tcp	10.6.109.62:8800*load*			map[]
```
[...]

Issue #268, opened 2017-02-21
Papertrail stops gettings logs after an hour


tleyden:
I'm running logspout on docker cloud with a volume mount of `-v /var/run/docker.sock:/var/run/docker.sock` and a command line arg to point to `syslog://logs5.papertrailapp.com:3...`

It runs for about an hour and then logs mysteriously stop appearing in papertrail.  

Is there anything I can do to debug it?

tleyden:
I should mention that I'm nowhere close to hitting the papertrail limit

grydstedt:
Seeing this as well. Any resolution @tleyden?

tleyden:
@grydstedt Unfortunately no.  Docker support told me to try a bigger instance, which I haven't done (partly because it just seems to sweep the issue under the carpet).

timstallmann:
Think I may be seeing something similar here as well :/

grydstedt:
Yeah, works for a couple of days of logging and then stops. I reached out to papertrail and they suggested these potential workarounds: 

- Make sure you're on the latest logspout. Currently, that's v3.1. If that doesn't fix it, it may be worth trying logspout:master, even though that's a prerelease.
- Point logspout at a local rsyslog server, which is configured to relay to Papertrail.
- Use Docker's built-in syslog driver instead of logspout. This requires access to the host, so that you can run dockerd with options like --log-driver=syslog and --log-opt syslog-address=tcp+tls://logsN.papertrailapp.com:99999. Also, it's got a downside that logs are no longer stored locally, even for the docker logs command – they're sent to syslog and that's all.

I'm already running logspout:master, so I might give setting up a rsyslog daemon a whirl

michaelshobbs:
We have a couple PRs open on this. I am looking at testing them and hopefully releasing a new image soon

michaelshobbs:
potentially related to #277

luketurner:
I've also encountered this problem, and it was because Papertrail resets TCP connections after 15 minutes of inactivity. Go considers an ECONNRESET to be a "temporary" error, so the logspout does not try to recreate the connection, but instead keeps retrying sending data on the broken connection.

In terms of solutions, I'm using a logspout patch that makes ECONNRESET a permanent error (because it seems silly for it to be temporary). I also have it send the logs to a local rsyslog container, which is responsible for queueing logs and retrying if Papertrail goes down. Overall, this scheme's dramatically increased the stability of my logs.

josegonzalez:
[...]

PR #253, opened 2017-01-06
Reconnect when syslog connection is failed.

roman-vynar:
Adresses https://github.com/gliderlabs/logspout/issues/215, https://github.com/gliderlabs/logspout/issues/107

Currently, when syslog connection is failed, logspout tries 11 times and gives up (it would have, see the last paragraph). Thus, it requires restart. This PR makes it to reconnect indefinitely every 2s.

Also do not close message channel on timeout 1s (defer function). This is also a case when syslog connection failed. When we restore connection, the closed channel prevents messages to come through.

Issue #215, opened 2016-07-28
logspout fails to reconnect after connection reset by peer.


jeacott1:
I have 5 instances of logspout running in marathon talking to loggly.
all was well until yesterday when all 5 failed with errors like this:

```
#   ADAPTER     ADDRESS                 CONTAINERS      SOURCES OPTIONS
#   syslog+tcp  logs-01.loggly.com:514                          map[]
2016/07/09 03:39:42 syslog: write tcp xxx.xxx.xxx.xxx:34657->logglyip:514: write: connection reset by peer
2016/07/09 03:39:42 syslog: retrying tcp up to 11 times
2016/07/09 03:40:23 syslog: retry failed
2016/07/09 03:40:23 syslog: reconnecting up to 11 times
2016/07/10 23:33:50 syslog: write tcp xxx.xxx.xxx.xxx:34657->logglyip:514: write: connection timed out
2016/07/10 23:33:50 syslog: retrying tcp up to 11 times
2016/07/10 23:34:31 syslog: retry failed
2016/07/10 23:34:31 syslog: reconnecting up to 11 times
2016/07/11 02:04:47 syslog: write tcp xxx.xxx.xxx.xxx:34657->logglyip:514: write: connection timed out
2016/07/11 02:04:47 syslog: retrying tcp up to 11 times
2016/07/11 02:05:28 syslog: retry failed
2016/07/11 02:05:28 syslog: reconnecting up to 11 times
2016/07/11 02:32:28 syslog: write tcp xxx.xxx.xxx.xxx:34657->logglyip:514: write: connection timed out
2016/07/11 02:32:28 syslog: retrying tcp up to 11 times
2016/07/11 02:33:09 syslog: retry failed
2016/07/11 02:33:09 syslog: reconnecting up to 11 times
2016/07/12 02:57:56 syslog: write tcp xxx.xxx.xxx.xxx:34657->logglyip:514: write: connection timed out
2016/07/12 02:57:56 syslog: retrying tcp up to 11 times
2016/07/12 02:58:37 syslog: retry failed
2016/07/12 02:58:37 syslog: reconnecting up to 11 times
2016/07/13 23:51:10 syslog: write tcp xxx.xxx.xxx.xxx:34657->logglyip:514: write: broken pipe
2016/07/13 23:51:10 syslog: reconnecting up to 11 times
2016/07/18 11:11:16 syslog: write tcp xxx.xxx.xxx.xxx:34657->logglyip:514: write: broken pipe
[...]

PR #291, opened 2017-04-12
preserve buffer on reconnect() AND reconnect on connection reset by peer

michaelshobbs:
- includes reconnect test
- disables the race detector on alpine but you can still run `make test-direct` locally with it on. ref: https://github.com/golang/go/issues/14481

turns out that in linux (alpine at least) a `conn.Close()` shows up as a `syscall.ECONNRESET`. so i included #287 here as well

closes #287 #277 #268 #253 #215 #107


adapters/syslog/syslog_test.go
```diff
@@ -1,22 +1,210 @@
...
+		t.Errorf("bad transport: " + route.Adapter)
+	}
+
+	ch := make(chan error, 1)
+	dchan := make(chan []byte, maxMsgCount)
+
+	// Read from connection
+	go func() {
+		defer close(ch)
+		defer close(dchan)
```
alif: nitpick
```
defer func() {
   close(ch)
   close(chan)
} ()
```

adapters/syslog/syslog_test.go
```diff
@@ -1,22 +1,210 @@
...
+		tmpl:      tmpl,
+		transport: transport,
+	}
+
+	// Send logstream
+	logstream := make(chan *router.Message)
+	done := make(chan bool)
+	go func() {
+		defer close(logstream)
+		defer close(done)
```
alif: similar nitpick on grouping defer into `func`

adapters/syslog/syslog_test.go
```diff
@@ -1,22 +1,210 @@
...
 	}
+
+	os.Unsetenv("RETRY_COUNT")
+	setRetryCount()
+	if retryCount != defaultRetryCount {
+		t.Errorf("expected %v got %v", defaultRetryCount, retryCount)
+	}
+}
+
+func TestSyslogReconnectOnClose(t *testing.T) {
```
alif: This test is pretty intense.
I'd recommend breaking it down a little. Maybe move the `go func()` to named functions so they're just `go someReallyClearFuncName()`
Would make things a little more readable.

michaelshobbs: done

adapters/syslog/syslog_test.go
```diff
@@ -1,22 +1,216 @@
...
+		Name: "0michaelshobbs",
+		Config: &docker.Config{
+			Hostname: "8dfafdbc3a40",
+		},
 	}
+	testPriority  = "{{.Priority}}"
+	testTimestamp = "{{.Timestamp}}"
+	testHostname  = "{{.Container.Config.Hostname}}"
+	testTag       = "{{.ContainerName}}"
+	testPid       = "{{.Container.State.Pid}}"
```
alif: Seems like L31-35 could be `const`

adapters/syslog/syslog_test.go
```diff
@@ -1,22 +1,216 @@
...
+
+	for err := range errc {
+		t.Errorf("%v", err)
+	}
+
+	for {
+		select {
+		case <-done:
+			if maxMsgCount-1 != len(datac) {
+				t.Errorf("expected %v got %v", maxMsgCount-1, len(datac))
```
mattaitchison: This should check for duplicate messages in addition to msg count.

michaelshobbs: done

adapters/syslog/syslog_test.go
michaelshobbs: done

testutil/mockconn.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+	if err := c.Client.Close(); err != nil {
+		return err
+	}
+	return nil
+}
+
+// NewConn returns a new testutil.Conn
+func NewConn() *Conn {
+	// A connection consists of two pipes:
```
progrium: Since a net.Pipe is a ReadWriter, it represents a duplex channel. I'm not sure you need two of these, which means you probably don't need a lot of the other code you have around these.

testutil/testutil.go
```diff
@@ -0,0 +1,98 @@
...
+// Test constants
+const (
+	CloseOnMsgIdx = 5
+	MaxMsgCount   = 10
+)
+
+// LocalTCPServer tcp server for testing specific network errors
+type LocalTCPServer struct {
+	lnmu     sync.RWMutex
+	MockConn *Conn
```
progrium: The approach of mocking net connections is actually not terribly idiomatic. Because it's relatively easy in Go, most network tests in Go actually stand up a full network client/server instead of building mocks. This reduces a lot of code and better represents reality.

michaelshobbs: I agree. If we can force specific connection errors (i.e. timeouts/resets/etc) then we can abandon this approach for sure. I'm playing with some stuff @mattaitchison sent over to see if we can use a "normal" `net.Conn` object.

progrium: We should be able to. And specific errors are specific cases for specific tests, which can minimally mock what's necessary.

michaelshobbs: That's a lot of should happening here. Feel free to chime in with specific pointers.....

michaelshobbs:
@progrium @mattaitchison updated this per our discussion

tleyden:
When is this expected to get pushed up to `gliderlabs/logspout:latest` in dockerhub?

michaelshobbs:
This is v3.2.1 and latest now

Issue #216, opened 2016-07-28
Logspout stops logging after truncating a logfile from a container


SqiSch:
Logspout stops logging after truncating a logfile from a container

to reproduce this, follow these steps: 

start a container f.e.:
`docker run -d ubuntu /bin/sh -c "while true; do echo hello world; sleep 1; done"`

check the logging via logspout: 
`curl localhost:7477/logs`

then truncate the logfile from the hello world container:
`truncate -s0 /data/docker/containers/zzzzz/yyyyyyy-json.log`

check the logging via logspout again: 
`curl localhost:7477/logs`

To fix the logging we need to restart logspout OR the hello world container, but the hello world container stills logging into the logfile.json

furkanayhan:
any update on this issue? @SqiSch

sybrandy:
I saw something similar today where I started logspout after some other containers and didn't get any logs.  I restarted the one container and logs started to come through.

furkanayhan:
For those who are dealing with this issue:
We are trying this tool now: https://sematext.com/blog/2016/04/04/docker-cloud-monitoring-logging/
And it seems to be working.

ebr:
we started running into this recently, and i suspect there was a change in how the docker daemon rotates the logs in v1.12, though I cannot find anything to support this in the v1.12 release notes. this is a big deal because we can't turn off log rotation.

ebr:
this may be the reason:
https://github.com/docker/docker/issues/23913

and this might be the fix:
https://github.com/docker/docker/pull/24514

ebr:
FWIW, we switched to `journald` log driver in docker (from `json-file`), and this is no longer an issue, because there's no file to be truncated. If you are stuck using `json-file` driver, this is definitely an issue. I actually had to create a "helper" service that would restart all logspout containers hourly as a workaround (and of course there was still a potential of missing up to 1 hour of logs, in case the truncation happened right after a restart)

mfornasa:
Any update here? We are experiencing this issue on Docker Cloud. Any workaround?

sybrandy:
The suggestion by @ebr to use journald appears to be working for me in some testing I was doing.  If you can do that, I'd try that first.

mfornasa:
@sybrandy my problem is that I am on Docker Cloud. They don't support docker log drivers.

olivergrahl:
[...]

PR #302, opened 2017-04-19
router/pump: remove logstream send timeout

mattaitchison:
Previously a containers logstream would be closed after a send on
its logstream channel was blocked for 1 second, causing it to not
forward logs until logspout received a start event for the container.

A send to a logstream could happen due to a closed syslog/tcp connection
or any delay in an adapters Stream from processing incoming messages.

Although this commit fixes the current issue, it does open up the
possibility for a send on a logstream to block for an indefinite amount
of time.

### DIFF
--- a/router/pump.go
+++ b/router/pump.go
@@ -371,15 +371,7 @@ func (cp *containerPump) send(msg *Message) {
 		if !route.MatchMessage(msg) {
 			continue
 		}
-		select {
-		case logstream <- msg:
-		case <-time.After(time.Second * 1):
-			debug("pump.send(): send timeout, closing")
-			// normal call to remove() triggered by
-			// route.Closer() may not be able to grab
-			// lock under heavy load, so we delete here
-			defer delete(cp.logstreams, logstream)
-		}
+		logstream <- msg
 	}
 }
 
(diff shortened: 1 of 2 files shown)


router/pump.go
```diff
@@ -373,12 +373,12 @@ func (cp *containerPump) send(msg *Message) {
 		}
 		select {
 		case logstream <- msg:
-		case <-time.After(time.Second * 1):
-			debug("pump.send(): send timeout, closing")
+			// case <-time.After(time.Second * 1):
```
michaelshobbs: Perhaps we should just rm this block since we can retrieve it from git history.

michaelshobbs:
Are there issues that should be linked?

mattaitchison:
Anything to do with lost logs could be linked.

michaelshobbs:
Should we maybe make `cp.logstreams` a buffered channel to *attempt* to smooth over any adapter issues?

mattaitchison:
We could but what is an ideal size?? 

I think the real solution would involve refactoring stream adapters and the log pump to not use channels or at least hide them from the public api. I have a few ideas but I will have to play with them a little more before implementing or even spec'ing them out.

michaelshobbs:
That's a fair question. Something bigger than `0` to start? Either way, though, is this something folks are seeing out in the wild? I couldn't find any open issues this seems to solve but maybe I missed something.

michaelshobbs:
possibly related to #216

PR #303, opened 2017-04-29
187 repro case

vignesh-kumar19:


michaelshobbs:
A solution has already been accepted and released to address #187. Please give it a shot in the `:latest` image and let us know if something is still not working correctly.

PR #307, opened 2017-05-24
Close existing routes when adding a new route with an existing ID.

### LINKED ISSUE #305: Can add duplicate routes with the same ID - undeletable
Start a ping so some logs are being generated.

    docker run alpine ping google.com

Create a logspout container.

    docker run -d -v /var/run/docker.sock:/var/run/docker.sock -p 8123:80 gliderlabs/logspout:master

Add a route to it via a POST REST interface with the id "abc":

    curl -X POST --data '{"id":"abc", "address":"logs3.papertrailapp.com:YOURPORTHERE", "adapter": "syslog"}' http://localhost:8123/routes

Observe logs are arriving correctly:
```
May 24 11:46:13 e30cb0feaa1a romantic_mahavira:  64 bytes from 216.58.208.142: seq=430 ttl=53 time=13.333 ms 
May 24 11:46:14 e30cb0feaa1a romantic_mahavira:  64 bytes from 216.58.208.142: seq=431 ttl=53 time=13.402 ms 
May 24 11:46:15 e30cb0feaa1a romantic_mahavira:  64 bytes from 216.58.208.142: seq=432 ttl=53 time=13.552 ms 
```
Run the CURL command again with the same ID:

    curl -X POST --data '{"id":"abc", "address":"logs3.papertrailapp.com:YOURPORTHERE", "adapter": "syslog"}' http://localhost:8123/routes
Observe that log lines are now being duplicated and sent twice (IMO **Incorrect behaviour**):
```
May 24 11:46:49 e30cb0feaa1a romantic_mahavira:  64 bytes from 216.58.208.142: seq=466 ttl=53 time=14.168 ms 
May 24 11:46:49 e30cb0feaa1a romantic_mahavira:  64 bytes from 216.58.208.142: seq=466 ttl=53 time=14.168 ms 
May 24 11:46:50 e30cb0feaa1a romantic_mahavira:  64 bytes from 216.58.208.142: seq=467 ttl=53 time=13.143 ms 
May 24 11:46:50 e30cb0feaa1a romantic_mahavira:  64 bytes from 216.58.208.142: seq=467 ttl=53 time=13.143 ms 
May 24 11:46:51 e30cb0feaa1a romantic_mahavira:  64 bytes from 216.58.208.142: seq=468 ttl=53 time=13.144 ms 
May 24 11:46:51 e30cb0feaa1a romantic_mahavira:  64 bytes from 216.58.208.142: seq=468 ttl=53 time=13.144 ms 
```

A call to /routes only shows a single route:
    curl localhost:8123/routes
```
[
  {
    "id": "abc",
    "adapter": "syslog",
    "address": "logs3.papertrailapp.com:24818"
  }
]
```

The /routes volume only shows the route once:
```
$ sudo ls /var/lib/docker/volumes/e1ca17038e88350a3ede82092c04dc46003bea32bb418709e50ad5cba686785d/_data
abc.json
$ sudo cat /var/lib/docker/volumes/e1ca17038e88350a3ede82092c04dc46003bea32bb418709e50ad5cba686785d/_data/abc.json
{
  "id": "abc",
  "adapter": "syslog",
  "address": "logs3.papertrailapp.com:24818"
}
```

This also "hides" the old route. If you do a DELETE /routes/myId, only the latest route is deleted, the old ones continue to keep sending.
```
curl -X DELETE localhost:8123/routes/abc
curl localhost:8123/routes
[]
```

```
May 24 11:53:33 e30cb0feaa1a romantic_mahavira:  64 bytes from 216.58.208.142: seq=870 ttl=53 time=13.269 ms 
May 24 11:53:34 e30cb0feaa1a romantic_mahavira:  64 bytes from 216.58.208.142: seq=871 ttl=53 time=14.185 ms 
May 24 11:53:35 e30cb0feaa1a romantic_mahavira:  64 bytes from 216.58.208.142: seq=872 ttl=53 time=13.184 ms 
May 24 11:53:36 e30cb0feaa1a romantic_mahavira:  64 bytes from 216.58.208.142: seq=873 ttl=53 time=13.382 ms
```

```
curl -X DELETE localhost:8123/routes/abc
404 page not found
```

I can't see a way to delete this "phantom" route. IMO the correct behavior should be to replace the route if a route is added with an existing ID (or reject the second route add, but I prefer replacing it as that fits my use-case where I want to update the route)

Happens on latest master (eb9bd8435f68b8cdef560fc4fb334858cb8a3321).

Could be the cause of #232 ?

Crashthatch:
Fixes #305 .

I added a test, and then needed to add a closed flag to verify that closer was being called. If anyone has a better way to do this, that doesn't require the extra flag I'd be interested to see it!

### DIFF
--- a/router/routes.go
+++ b/router/routes.go
@@ -136,6 +136,11 @@ func (rm *RouteManager) Add(route *Route) error {
 	}
 	route.closer = make(chan bool)
 	route.adapter = adapter
+	//Stop any existing route with this ID:
+	if( rm.routes[route.ID] != nil ){
+		rm.routes[route.ID].closer <- true
+	}
+
 	rm.routes[route.ID] = route
 	if rm.persistor != nil {
 		if err := rm.persistor.Add(route); err != nil {
--- a/router/routes.go
+++ b/router/routes.go
@@ -137,7 +137,7 @@ func (rm *RouteManager) Add(route *Route) error {
 	route.closer = make(chan bool)
 	route.adapter = adapter
 	//Stop any existing route with this ID:
-	if( rm.routes[route.ID] != nil ){
+	if rm.routes[route.ID] != nil {
 		rm.routes[route.ID].closer <- true
 	}
 
--- a/router/routes_test.go
+++ b/router/routes_test.go
@@ -5,6 +5,18 @@ import (
 	"testing"
 )
 
+type DummyAdapter struct{}
+
+func (a *DummyAdapter) Stream(logstream chan *Message) {
+	for message := range logstream {
+		print( "message passed to DummyAdapter: ", message )
+	}
+}
+
+func newDummyAdapter(route *Route) (LogAdapter, error) {
+	return &DummyAdapter{}, nil
+}
+
 func TestRouterGetAll(t *testing.T) {
 	rts, err := Routes.GetAll()
 	if err != nil {
@@ -14,5 +26,35 @@ func TestRouterGetAll(t *testing.T) {
 	emptyRoutes := append(marshal(make([]*Route, 0)), '\n')
 	if !reflect.DeepEqual(routes, emptyRoutes) {
 		t.Error("expected '[]' got:", routes)
+	}
+}
+
+func TestRouterNoDuplicateIds(t *testing.T) {
+	AdapterFactories.Register(newDummyAdapter, "syslog")
+
+	//Mock "running" so routes actually start running when added.
+	Routes.routing = true
+
+	//Start the first route.
+	route1 := &Route{
+		ID: "abc",
+		Address: "someUrl",
+		Adapter: "syslog",
+	}
+	err := Routes.Add(route1)
+	if err != nil {
+		t.Error("Error adding route:", err)
+	}
+
+	//Start a second route with the same ID.
+	var route2 = &Route{
+		ID: "abc",
+		Address: "someUrl2",
+		Adapter: "syslog",
+	}
+	Routes.Add(route2)
+
+	if( !route1.closed ){
+		t.Errorf("route1 was not closed after route2 added." )
 	}
 }
--- a/router/routes_test.go
+++ b/router/routes_test.go
@@ -9,7 +9,7 @@ type DummyAdapter struct{}
 
 func (a *DummyAdapter) Stream(logstream chan *Message) {
 	for message := range logstream {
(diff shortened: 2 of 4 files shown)


router/routes.go
```diff
@@ -136,6 +136,11 @@ func (rm *RouteManager) Add(route *Route) error {
 	}
 	route.closer = make(chan bool)
 	route.adapter = adapter
+	//Stop any existing route with this ID:
+	if( rm.routes[route.ID] != nil ){
```
michaelshobbs: nitpick: `()` are unnecessary

router/routes_test.go
```diff
@@ -16,3 +28,33 @@ func TestRouterGetAll(t *testing.T) {
...
+	//Mock "running" so routes actually start running when added.
+	Routes.routing = true
+
+	//Start the first route.
+	route1 := &Route{
+		ID: "abc",
+		Address: "someUrl",
+		Adapter: "syslog",
+	}
+	err := Routes.Add(route1)
```
michaelshobbs: nitpick: you can combine L44 & L45
`if err := Routes.Add(route1); err != nil {`

router/routes_test.go
```diff
@@ -16,3 +28,33 @@ func TestRouterGetAll(t *testing.T) {
...
+
+	//Start a second route with the same ID.
+	var route2 = &Route{
+		ID: "abc",
+		Address: "someUrl2",
+		Adapter: "syslog",
+	}
+	Routes.Add(route2)
+
+	if( !route1.closed ){
```
michaelshobbs: ditto on `()`

router/routes_test.go
```diff
@@ -16,3 +28,33 @@ func TestRouterGetAll(t *testing.T) {
...
+		t.Errorf("route1 was not closed after route2 added." )
+	}
+}
```
michaelshobbs: missing newline

router/routes.go
michaelshobbs: This question might be naive but should we throw an error instead of overwriting the route?

Crashthatch: @michaelshobbs I did wonder about throwing an error instead of overwriting (see #307 ). In my use case I need to replace the route, ideally without losing any log lines (which could happen if it were stopped through one HTTP call and started it through another), so I did it this way.

Do others have route-replace use-cases / opinions?

Issue #246, opened 2016-12-13
CPU usage on dockerd 100%


runningman84:
If I start logspout-logstash using this docker-compose the cpu usage spikes to 100% and stays there...

```
  logspout:
    image: bekt/logspout-logstash
    mem_limit: 256m
    restart: on-failure
    depends_on:
      - logstash
    environment:
      ROUTE_URIS: logstash://logstash:5000
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
```
before starting logspout
```
 7015 root      20   0 1423204  80956  29440 S 106.3  1.3   1:25.72 dockerd       
```
after starting logspout
```
 7015 root      20   0 1423204  80972  29440 S   0.3  1.3   1:58.57 dockerd  
```

pythianarceri:
Experiencing the same issue here:


# logspout --version                                                          
v3.2-dev-custom   

# docker version
Client:
 Version:      1.12.3
 API version:  1.24
 Go version:   go1.6.3
 Git commit:   6b644ec
 Built:        Wed Oct 26 23:26:11 2016
 OS/Arch:      linux/amd64

Server:
 Version:      1.12.3
 API version:  1.24
 Go version:   go1.6.3
 Git commit:   6b644ec
 Built:        Wed Oct 26 23:26:11 2016
 OS/Arch:      linux/amd64

Logspout container logs with DEBUG enabled show a constant stream of:
1/26/2017 12:37:44 PM2017/01/26 17:37:44 pump.pumpLogs(): 495069bafc93 stopped
1/26/2017 12:37:45 PM2017/01/26 17:37:45 pump.pumpLogs(): 495069bafc93 started
1/26/2017 12:37:45 PM2017/01/26 17:37:45 pump.pumpLogs(): 25635c95f804 stopped
1/26/2017 12:37:45 PM2017/01/26 17:37:45 pump.pumpLogs(): 25635c95f804 started
1/26/2017 12:37:46 PM2017/01/26 17:37:46 pump.pumpLogs(): 495069bafc93 stopped
1/26/2017 12:37:46 PM2017/01/26 17:37:46 pump.pumpLogs(): 495069bafc93 started
1/26/2017 12:37:46 PM2017/01/26 17:37:46 pump.pumpLogs(): a27d4bfb3a41 stopped
1/26/2017 12:37:46 PM2017/01/26 17:37:46 pump.pumpLogs(): a27d4bfb3a41 started
[...]

PR #321, opened 2017-09-01
Tail env

stevecalvert:
Add a new environment variable, "TAIL" to allow the docker logs --tail option to be configured.  At present, it is hardcoded to "all"

Setting TAIL=0 looks like a workaround for issue #246 since it avoids parsing of the earlier and potentially corrupt content of the JSON log file.  Note, at present, the BACKLOG=false option does not workaround this issue since the logfile is still parsed to determine the timestamps of the logs

Update the README with details of the TAIL envvar

### DIFF
--- a/router/pump.go
+++ b/router/pump.go
@@ -195,6 +195,7 @@ func (p *LogsPump) pumpLogs(event *docker.APIEvents, backlog bool, inactivityTim
 		return
 	}
 
+	var tail = getopt("TAIL", "all")
 	var sinceTime time.Time
 	if backlog {
 		sinceTime = time.Unix(0, 0)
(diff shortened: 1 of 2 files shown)


router/pump.go
```diff
@@ -215,18 +215,18 @@ func (p *LogsPump) pumpLogs(event *docker.APIEvents, backlog bool, inactivityTim
 	p.update(event)
 	go func() {
 		for {
-			debug("pump.pumpLogs():", id, "started")
+			debug("pump.pumpLogs():", id, "started, tail:", getopt("TAIL", "all"))
```
michaelshobbs: i'd rather see `getopt` called once. can you update this to set a variable use that in these two places?

stevecalvert: yes, fair point - updated

michaelshobbs:
@stevecalvert have you used this live yet?

stevecalvert:
changes have been tested locally against my branch.  Note, I've also updated the README to add TAIL to the lis of environment vars  - I overlooked that on the first commit

PR #325, opened 2017-09-20
Export a metrics endpoint for monitoring

pegerto:
I find difficult to understand how logspout is behaving, one important metric for me is the rate of the sent msg per container. So we can identify peaks on the clusters easily but many other metrics can be really useful. 

This PR expose a metric endpoint for prometheus that I can use to monitor logspout across my k8s cluster. 

Any feedback will be more than welcome.


pegerto:
I see the testing issue,
```
 go list ./... | egrep -v 'custom|vendor'
```

Will install the package dependencies for the master branch, prometheus as a dependency, so its is a bit chicken and egg problem

router/metrics.go
```diff
@@ -0,0 +1,21 @@
...
+	"net/http"
+)
+
+// Pump message send counter
+var pumpMsgSend *prometheus.CounterVec
+
+func init() {
+	pumpMsgSend = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "pump_msg_send",
```
discordianfish: All prometheus counter metrics should end in `_total`. Personally would call it `logstash_sent_messages_total` or `logstash_messages_sent_total`.

discordianfish:
@progrium is this something you are willing to merge in general?

I can help reviving this PR. There is probably a better way than initializing all this in init().

Issue #328, opened 2017-09-26
Would need a way to specify the hostname to use when running logspout as a global service


chris7444:
issue #181 describes what I want to achieve with logspout
[https://github.com/gliderlabs/logspout/issues/181](url)

To summarize, here is how I want logspout to output messages

Format1:  timestamp  field2 NameOfContainer ...

where field2 is the name of the docker host running the container rather than an ID (default)

I know I can use the SYSLOG_HOSTNAME environment variable to have logspout spit messages with the name of the docker host but this does not work when defining a global service for logspout

example compose file

version: "3"
networks:
  logging:
services:
  logspout:
    image: gliderlabs/logspout:{{ logspout_version }}
    environment:
      -SYSLOG_HOSTNAME="something"
    networks:
      - logging
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
    command:
      syslog://{{groups.logger[0]}}.{{domain_name}}:514
    deploy:
      mode: global

The value for SYSLOG_HOSTNAME will be the same for all replicas of the logspout container and this is not what I want. 

A suggestion would be to do something similar to what is done for node-exporter . In the following compose file, I can see that the name of the host is passed to the container by mounting /etc/hostname to /etc/host_hostname. By using a similar technic it would be possible to run logspout as a global service and have each and every instance 

  node-exporter:
    image: basi/node-exporter:{{ node_exporter_version }}
    networks:
      - monitoring
    ports:
      - 9100:9100
    volumes:
      - /proc:/host/proc:ro
      - /sys:/host/sys:ro
      - /:/rootfs:ro
      - /etc/hostname:/etc/host_hostname:ro
    environment:
      - HOST_HOSTNAME=/etc/host_hostname

more info  here: [https://hub.docker.com/r/basi/node-exporter/](url)

chris7444:
suggested change
(please forgive me I am not using a pull request now but I am learning everything at the same time, docker, ansible, python, go, git...)

[root@svt2-ucp01 logspout]# git diff adapters/syslog/syslog.go
diff --git a/adapters/syslog/syslog.go b/adapters/syslog/syslog.go
index e44f787..3b8b27c 100644
--- a/adapters/syslog/syslog.go
+++ b/adapters/syslog/syslog.go
@@ -4,6 +4,7 @@ import (
        "bytes"
        "errors"
        "fmt"
+       "io/ioutil"
        "log"
        "log/syslog"
        "net"
[...]

PR #329, opened 2017-09-26
take the hostname from /etc/host_hostname if the file is there. This …

chris7444:
using environment variable SYSLOG_HOSTNAME does not help when  logspout is used as a a global service
this changes consists in looking for the file /etc/host_hostname to find the SYSLOG_HOSTNAME
typically, one will run the container with -v/etc/hostname:/etc/host_hostname

When using a compose file, the file will look like something like this

version: "3"

networks:
  logging:

services:
  logspout:
    image: svt2-dtr.am2.cloudra.local/team/logspout:latest
    environment:
      - ALLOW_TTY=1
    networks:
      - logging
    volumes:
      - /etc/hostname:/etc/host_hostname:ro
      - /var/run/docker.sock:/var/run/docker.sock
    command:
      syslog://svt2-logger.am2.cloudra.local:514
    deploy:
      mode: global
      resources:
        limits:
          cpus: '0.20'
          memory: 256M
        reservations:
          cpus: '0.10'
          memory: 128M

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ Logspout relies on the Docker API to retrieve container logs. A failure in the A
 
 #### Using Logspout in a swarm
 
-In a swarm, logspout is best deployed as a global service.  When running logspout with 'docker run', you can change the value of the hostname field using the SYSLOG_HOSTNAME environment variable as explained above. However, this does not work in a compose file because the value for SYSLOG_HOSTNAME will be the same for all logspout "tasks", regardless of the docker host on which they run. To support this mode of deployment, the syslog adapter will look for the file /etc/host_hostname and, if the file exists and it is not empty, will configure the hostname field with the content of this file. You can then use a volume mount to map a file on the docker hosts with the file /etc/host_hostname in the container.  The sample compose file below illustrates how this can be done
+In a swarm, logspout is best deployed as a global service.  When running logspout with 'docker run', you can change the value of the hostname field using the `SYSLOG_HOSTNAME` environment variable as explained above. However, this does not work in a compose file because the value for `SYSLOG_HOSTNAME` will be the same for all logspout "tasks", regardless of the docker host on which they run. To support this mode of deployment, the syslog adapter will look for the file `/etc/host_hostname` and, if the file exists and it is not empty, will configure the hostname field with the content of this file. You can then use a volume mount to map a file on the docker hosts with the file `/etc/host_hostname` in the container.  The sample compose file below illustrates how this can be done
 
 ```
 version: "3"
--- a/README.md
+++ b/README.md
@@ -59,17 +59,17 @@ You can tell logspout to only include certain containers by setting filter param
 		--volume=/var/run/docker.sock:/var/run/docker.sock \
 		gliderlabs/logspout \
 		raw://192.168.10.10:5000?filter.name=*_db
-
+	
 	$ docker run \
 		--volume=/var/run/docker.sock:/var/run/docker.sock \
 		gliderlabs/logspout \
 		raw://192.168.10.10:5000?filter.id=3b6ba57db54a
-
+	
 	$ docker run \
 		--volume=/var/run/docker.sock:/var/run/docker.sock \
 		gliderlabs/logspout \
 		raw://192.168.10.10:5000?filter.sources=stdout%2Cstderr
-
+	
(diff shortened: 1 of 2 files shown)


michaelshobbs:
closes #328

adapters/syslog/syslog.go
```diff
@@ -67,8 +68,14 @@ func NewSyslogAdapter(route *router.Route) (router.LogAdapter, error) {
 
 	format := getopt("SYSLOG_FORMAT", "rfc5424")
 	priority := getopt("SYSLOG_PRIORITY", "{{.Priority}}")
-	hostname := getopt("SYSLOG_HOSTNAME", "{{.Container.Config.Hostname}}")
 	pid := getopt("SYSLOG_PID", "{{.Container.State.Pid}}")
+
+	b, err := ioutil.ReadFile("/etc/host_hostname") // just pass the file name
+	if err != nil {
```
michaelshobbs: Should also check for empty `b` here.

michaelshobbs:
@chris7444 to clarify here, "global service" means a **docker service** running in a swarm cluster, correct?

chris7444:
@michaelshobbs global service means there will be one instance of the service (called a task) running on each and every  docker node in the swarm. Which is a perfect fit for monitoring agents or containers like logspout.  The moment you add a node in the swarm, docker will make sure one replica of the service will run on the new node. The other deployment option is "replicated" where you say I want 'N' instances of the service running in the swarm (eg 3 nginx instances). The swarm makes sure there will always be at least 3 instances of nginx running

README.md
```diff
@@ -156,6 +156,44 @@ Logspout relies on the Docker API to retrieve container logs. A failure in the A
 * `SYSLOG_TAG` - datum for tag field (default `{{.ContainerName}}+route.Options["append_tag"]`)
 * `SYSLOG_TIMESTAMP` - datum for timestamp field (default `{{.Timestamp}}`)
 
+#### Using Logspout in a swarm
+
+In a swarm, logspout is best deployed as a global service.  When running logspout with 'docker run', you can change the value of the hostname field using the SYSLOG_HOSTNAME environment variable as explained above. However, this does not work in a compose file because the value for SYSLOG_HOSTNAME will be the same for all logspout "tasks", regardless of the docker host on which they run. To support this mode of deployment, the syslog adapter will look for the file /etc/host_hostname and, if the file exists and it is not empty, will configure the hostname field with the content of this file. You can then use a volume mount to map a file on the docker hosts with the file /etc/host_hostname in the container.  The sample compose file below illustrates how this can be done
```
michaelshobbs: nitpick: place `SYSLOG_HOSTNAME` and `/etc/host_hostname` in backticks

chris7444:
@michaelshobbs . I did the changes in README.md and yes I did test the changes in my env (pulling the image from my repo)

chris7444:
@michaelshobbs . What is your release schedule. when can I expect to see the changes merged into your code ? Is there anything I need to do for this to happen ?

michaelshobbs:
@chris7444 just haven't had time to circle back. i'll try to cut a release today. otherwise this is now in master. give it about 30 minutes to show up on docker hub though

nvanheuverzwijn:
@michaelshobbs @chris7444 Using this feature breaks my logs. I use a service called "logdna" to aggregate all logs. When I mount /etc/host_hostname, the host is correctly displayed but the message is truncated after.

When I hardcode the hostname using SYSLOG_HOSTNAME, everything works fine.

Could it be possible that reading a file with go would add some hidden character ?

josegonzalez:
Probably you have a newline character, and we should strip whitespace from the file after its read.

nvanheuverzwijn:
@josegonzalez There is no newline, I really don't get it. There is no weird character or anything.

nvanheuverzwijn:
Any way to dump the raw content that is sent to the server ? Some kind of debug mode or something ?

josegonzalez:
You're not understanding me. Your file has two lines, and we're reading the whole thing. So the output includes a newline, which is probably causing logs being shipped to be truncated.

The fix is literally for something in golang to trim all whitespace from both ends. If you want to contribute that, that would be great.

nvanheuverzwijn:
Oh, you are right, I got confused because `vim` does not show all character :-((

I just checked my file with `cat /etc/hostname | od -c` and there was effectively a \n character at the end.

PR #333, opened 2017-10-10
Formatting logspout

nkangas:


michaelshobbs:
This seems like it was mistakenly opened against the upstream. Closing.

PR #335, opened 2017-10-12
strip \r and \n when reading the file /etc/host_hostname

nvanheuverzwijn:
related to #329

PR #338, opened 2017-10-13
strip \r and \n when reading the file /etc/host_hostname

michaelshobbs:
also:
- [x] fixes CI `working_directory` so we don't have duplicate packages
- [x] upgrade to `alpine:3.6` in order to get golang 1.8.3
- [x] pass in `DEBUG` to test container

closes #335

### DIFF
--- a/adapters/syslog/syslog.go
+++ b/adapters/syslog/syslog.go
@@ -56,6 +56,16 @@ func debug(v ...interface{}) {
 	}
 }
 
+func getHostname() string {
+	content, err := ioutil.ReadFile("/etc/host_hostname")
+	if err == nil && len(content) > 0 {
+		hostname = strings.TrimRight(string(content), "\r\n")
+	} else {
+		hostname = getopt("SYSLOG_HOSTNAME", "{{.Container.Config.Hostname}}")
+	}
+	return hostname
+}
+
 // NewSyslogAdapter returnas a configured syslog.Adapter
 func NewSyslogAdapter(route *router.Route) (router.LogAdapter, error) {
 	transport, found := router.AdapterTransports.Lookup(route.AdapterTransport("udp"))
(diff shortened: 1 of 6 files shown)


michaelshobbs:
@nvanheuverzwijn comments welcomed here

nvanheuverzwijn:
@michaelshobbs Thank you Michael, I appreciate your work greatly ! Have an excellent day.

adapters/syslog/syslog.go
```diff
@@ -55,6 +56,16 @@ func debug(v ...interface{}) {
 	}
 }
 
+func getHostname() string {
+	content, err := ioutil.ReadFile("/etc/host_hostname")
+	if err == nil && len(content) > 0 {
+		hostname = strings.TrimRight(string(content), "\r\n")
```
markdascher: This modifies the global hostname variable. Should we add a local declaration at the beginning of getHostname? (Or just return here instead of assigning the result?)

adapters/syslog/syslog.go
```diff
@@ -55,6 +56,16 @@ func debug(v ...interface{}) {
...
+	} else {
+		hostname = getopt("SYSLOG_HOSTNAME", "{{.Container.Config.Hostname}}")
```
markdascher: Same thing here.

adapters/syslog/syslog.go
```diff
@@ -69,13 +80,8 @@ func NewSyslogAdapter(route *router.Route) (router.LogAdapter, error) {
 	format := getopt("SYSLOG_FORMAT", "rfc5424")
 	priority := getopt("SYSLOG_PRIORITY", "{{.Priority}}")
 	pid := getopt("SYSLOG_PID", "{{.Container.State.Pid}}")
+	hostname = getHostname()
```
markdascher: I think this would make more sense as `:=` too.

josegonzalez:
@markdascher please file a new issue, rather than posting on an old pull request.

Issue #331, opened 2017-10-04
extra characters prepended to logs


gmelika:
After upgrading to 3.2.3 logspout started adding weird characters to the beginning of every line.  For example

```
�time="2017-10-04T01:20:14Z" level=info msg="Done checkpointing in-memory metrics and chunks in 1.404599334s." source="persistence.go:665"
xtime="2017-10-04T01:25:14Z" level=info msg="Checkpointing in-memory metrics and chunks..." source="persistence.go:633"
�time="2017-10-04T01:25:16Z" level=info msg="Done checkpointing in-memory metrics and chunks in 1.506158531s." source="persistence.go:665"
xtime="2017-10-04T01:30:16Z" level=info msg="Checkpointing in-memory metrics and chunks..." source="persistence.go:633"
�time="2017-10-04T01:30:17Z" level=info msg="Done checkpointing in-memory metrics and chunks in 1.52509546s." source="persistence.go:665"
xtime="2017-10-04T01:35:17Z" level=info msg="Checkpointing in-memory metrics and chunks..." source="persistence.go:633"
�time="2017-10-04T01:35:19Z" level=info msg="Done checkpointing in-memory metrics and chunks in 1.49991571s." source="persistence.go:665"
xtime="2017-10-04T01:40:19Z" level=info msg="Checkpointing in-memory metrics and chunks..." source="persistence.go:633"
�time="2017-10-04T01:40:20Z" level=info msg="Done checkpointing in-memory metrics and chunks in 1.417177673s." source="persistence.go:665"
xtime="2017-10-04T01:45:20Z" level=info msg="Checkpointing in-memory metrics and chunks..." source="persistence.go:633"
�time="2017-10-04T01:45:22Z" level=info msg="Done checkpointing in-memory metrics and chunks in 1.480621256s." source="persistence.go:665"
xtime="2017-10-04T01:50:22Z" level=info msg="Checkpointing in-memory metrics and chunks..." source="persistence.go:633"
�time="2017-10-04T01:50:23Z" level=info msg="Done checkpointing in-memory metrics and chunks in 1.516549895s." source="persistence.go:665"
xtime="2017-10-04T01:55:23Z" level=info msg="Checkpointing in-memory metrics and chunks..." source="persistence.go:633"
�time="2017-10-04T01:55:25Z" level=info msg="Done checkpointing in-memory metrics and chunks in 1.527579546s." source="persistence.go:665"
[...]

PR #346, opened 2017-10-30
Strip header bytes from log stream

jgreat:
Looks like docker decided to prepend an 8 byte header to each line in the log stream.  This is causing garbage characters to be added to the front of the message being pushed to the log server breaking json parsing for me.

Issue #331 shows some output of the problem.

This PR just discards the 8 bytes in the buffer. 

Not sure how you want the versioning.  Be happy to update this PR with any changes you need.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased][unreleased]
 ### Fixed
+- @jgreat Strip docker log header bytes from log stream. fixes #331
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,6 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased][unreleased]
 ### Fixed
-- @jgreat Strip docker log header bytes from log stream. fixes #331
 
 ### Added
 
--- a/VERSION
+++ b/VERSION
@@ -1,1 +1,1 @@
-v3.3-dev.jgreat.strip-log-headers
+v3.3-dev
--- a/VERSION
+++ b/VERSION
@@ -1,1 +1,1 @@
-v3.3-dev
+v3.3-dev.jgreat.strip-log-headers
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: 1b25e45c33a1a8a0afd97c2b43e30882e6fcd7f0eb82a136896a0fdbc30adf62
-updated: 2016-07-08T17:33:18.562222083-04:00
+hash: a061927dfe895ab75bdcccfa04e401d9c4bfe1f4f6d7301e3e8f571a5c06fc10
+updated: 2017-10-31T16:57:54.70566614-05:00
 imports:
 - name: github.com/docker/docker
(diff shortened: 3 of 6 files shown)


michaelshobbs:
closes #331

VERSION
```diff
@@ -1 +1 @@
-v3.3-dev
+v3.3-dev.jgreat.strip-log-headers
```
michaelshobbs: please revert this line

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased][unreleased]
 ### Fixed
+- @jgreat Strip docker log header bytes from log stream. fixes #331
```
michaelshobbs: no need to update change log. we do that as part of the release process. please revert

router/pump.go
```diff
@@ -347,6 +347,8 @@ func newContainerPump(container *docker.Container, stdout, stderr io.Reader) *co
 	pump := func(source string, input io.Reader) {
 		buf := bufio.NewReader(input)
 		for {
+			// Discard header bytes from log message
+			_, err := buf.Discard(8)
```
michaelshobbs: does it make sense to check the API version and only do this above the appropriate version?

jgreat: I've dug in a little deeper.  Looks like container/{id}/logs is following the same output pattern as the container/{id}/attach endpoint, with headers only being prepended when Tty=false. 

I've confirmed this behavior in docker 17.10.0-ce (API 1.33) and the oldest version I still have running 1.10.3 (API 1.22). From some of the GitHub issues it might have been set to do this since API 1.13 or before.

I'm going to update this and wrap the Discard in a check for > 1.13 API version and look for TTY=False.

References: 
https://docs.docker.com/engine/api/v1.33/#operation/ContainerAttach
https://github.com/moby/moby/issues/7375
https://github.com/moby/moby/issues/14201

glide.lock
```diff
@@ -30,6 +30,10 @@ imports:
   version: 9fa818a44c2bf1396a17f9d5a3c0f6dd39d2ff8e
 - name: github.com/hashicorp/go-cleanhttp
   version: ad28ea4487f05916463e2423a55166280e8254b5
+- name: github.com/hashicorp/go-version
+  version: fc61389e27c71d120f87031ca8c88a3428f372dd
+- name: github.com/looplab/logspout-logstash
```
michaelshobbs: i don't think we don't want this here

jgreat: I added the go-version lib to do the api version compare.

I think the logspout-logstash was added by glide since its in custom/modules.go file.  I'm not an expert on glide. How do I tell glide not to include that when I do glide get/install?

michaelshobbs: 👍 on the `go-version` lib

go ahead and manually remove L35-L36 here

jgreat:
Manually deleting the entries messes up the hash in the glide.lock file, so I added an exclude for the custom directory in the glide.yaml. This prevents the logstash module from being included in a `glide get`. 

`make build-custom` still builds the custom modules.

michaelshobbs:
Even better. I couldn’t remember the right lever to do that. Thanks!

Issue #352, opened 2017-11-08
Logspout + AWS ECS:


arnemohr:
With https://github.com/gliderlabs/logspout/commit/e0d92638b3f8e8eea4c5683498b9364b7e8e9050 you untroduced a "feature" which removes the first 8 Bytes from the log message.

router/pump.go
`
	if hasHeaders {
				buf.Discard(8)
	}
`

Unfortunately hasHeaders will always be true when DockerContainer are running on Amazon ECS (No TTY available when running container on ECS). Due to that logspout will always remove the first 8 bytes from my logstring, which breaks my jsonoutput. I imagine I'm not the only one who uses the combination container-engine + json message + elk stack.

PR #353, opened 2017-11-08
revert dropping header,  for only setting RawTerminal when container tty

jgreat:
This PR pretty much reverts the changes in PR #346 

After digging deeper into issue #331, it looks like actual issue is setting RawTerminal = allowTTY. 

Raw output is fine on containers that have Tty=true, but docker includes stream headers when container Tty=false.

### DIFF
--- a/router/pump.go
+++ b/router/pump.go
@@ -11,7 +11,6 @@ import (
 	"time"
 
 	"github.com/fsouza/go-dockerclient"
-	"github.com/hashicorp/go-version"
 )
 
 var allowTTY bool
@@ -181,10 +180,6 @@ func (p *LogsPump) Run() error {
 
 func (p *LogsPump) pumpLogs(event *docker.APIEvents, backlog bool, inactivityTimeout time.Duration) {
 	id := normalID(event.ID)
-
-	dockerVersion, err := p.client.Version()
-	assert(err, "Could not get Docker Version endpoint")
-
 	container, err := p.client.InspectContainer(id)
 	assert(err, "pump")
 	if ignoreContainerTTY(container) {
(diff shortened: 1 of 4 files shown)


jgreat:
#352

router/pump.go
```diff
@@ -214,9 +209,16 @@ func (p *LogsPump) pumpLogs(event *docker.APIEvents, backlog bool, inactivityTim
 		debug("pump.pumpLogs():", id, "pump exists")
 		return
 	}
+
+	// RawTerminal with contianer Tty=false injects binary headers into
```
michaelshobbs: nit: type `contianer`

jgreat: I'm not sure what you're asking for here?

jgreat:
https://github.com/moby/moby/issues/7375 - Describes the situation with headers showing up as unicode characters. This issue also describes the format of the headers and when they are injected.

https://github.com/moby/moby/issues/14201 - Complains about the lack of docs on the log stream.  Docker's response was "go look at the container/{id}/attach docs" :frowning_face:  

https://godoc.org/github.com/fsouza/go-dockerclient#Client.Logs - Talks about the difference in output if you are using LogsOptions.RawTerminal. 

There is a comment in the type LogsOptions for RawTerminal (https://github.com/fsouza/go-dockerclient/blob/master/container.go#L1472)  that says you "Usually true when the container contains a TTY"  Which I assume is why that was originally set to match allowTTY.

The problem is when your container Tty is false, and you set RawTerminal to true, you get the raw stream with headers.

So instead of just setting RawTerminal to true when allowTTY is true we really need to see if the container is Tty=true before using the RawTerminal.

jgreat:
The introduction of the RawTerminal option in v3.2.3 explains why the original reporter of #331 said that going back to v3.2.2 resolved the issue.

michaelshobbs:
Yea, was just looking for more context

router/pump.go
michaelshobbs: whoops I typo'd the word typo.

`container` is misspelled as  `contianer` in the comment

michaelshobbs:
Thanks for adding context here

PR #367, opened 2017-12-13
re-add execute permissions that were originally added in 3.2.3

djeeg:
to fix building on hosted docker environments

original discussion Oct 2016
https://github.com/gliderlabs/logspout/issues/238

re-emergence of issue Nov 2017
https://github.com/gliderlabs/logspout/issues/238#issuecomment-346996924


michaelshobbs:
accepted #363 only because i saw that before this 😄

PR #382, opened 2018-02-15
Sub-second accuracy in syslog adapter

### LINKED ISSUE #444: No milisecond on timestamps causes log ordering issues
Hello, I've commented to related PR chat but couldn't get answer so I'm asking from there.
https://github.com/gliderlabs/logspout/pull/401#issuecomment-544856706

Since log messages doesn't includes miliseconds in timestamps, observing exceptions is hard. Is there a parameter or something for getting timestamps from docker logs. It really hurts us and feels weird that only a few person talks about this issue, am I missing something?

romanlevin:
Other than unnecessarily losing timestamp accuracy, this breaks for syslog daemons that expect sub-second accuracy (like `fluentd`).


josegonzalez (commented):
Seems fine, do we need to have tests for this? Would it be better to somehow make this configurable?

romanlevin:
> Seems fine, do we need to have tests for this? Would it be better to somehow make this configurable?

@josegonzalez I'm new to contributing to logspout – are you asking me? I'm not super fluent in Go and the tests around the syslog are a bit of the mystery to me at the moment.

Making this opt-in with an environment variable should be super simple even for me, but I'm not sure if that's what you meant by "configurable".

jrd:
Could this be merged and release, please?

TomasLukl:
Could this be merged and release, please?

progrium:
Btw this LGTM ... let me know if repo needs to announce call for co-maintainers.

josegonzalez:
Yeah we should probably do that - I don't maintain the repo but Im sure there are folks that would want to.

TomasLukl (approved):

progrium (approved):

PR #384, opened 2018-03-08
Added handling for logs in JSON fromat in template

jozuenoon:
Strings or bytes formatted as proper JSON would not be marshaled another time. I added also informative error instead of "null" so any errors could be discovered and fixed.

### DIFF
--- a/adapters/raw/raw.go
+++ b/adapters/raw/raw.go
@@ -34,8 +34,8 @@ var funcs = template.FuncMap{
 
 		bytes, err := json.Marshal(value)
 		if err != nil {
-			log.Println("error marshalling to JSON: ", err)
-			return fmt.Sprintf("logspout: error marshalling to JSON: %s", err)
+			log.Println("error marshaling to JSON: ", err)
+			return fmt.Sprintf("logspout: error marshaling to JSON: %s", err)
 		}
 		return string(bytes)
 	},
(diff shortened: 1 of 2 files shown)


Dockerfile.dev
```diff
@@ -1,4 +1,4 @@
-FROM alpine:3.6
+FROM alpine:3.7
```
michaelshobbs: why this change?

masterada:
I don't think this is a good idea. This would mean that the structure of the log depends on it's content, eg:
RAW_FORMAT={"message":{{ toJSON .Data }}

Data1: `This is some error message.`
Data2: `{}`
Data3: `123`
Data4: `{ "debug-data": { "a": 2 } }`

It would be really painful to convert all these different formats back to string to strore them in a sigle message field. 

I would much rather see:
`{{if isJSON .Data}} .Data {{else}} toJSON .Data {{end}}`

Dockerfile.dev
jozuenoon: See commit comment. Alpine 3.7 brings go 1.9.4 with `json.Valid` function (afaik it's not available in Go version that was default in Alpine 3.6). But maybe there is other solution for that? Why it's important to stick with 3.6 anyway?

jozuenoon:
My changes would only prevent content that is already JSON formatted from being marshaled again (in my case this produces ugly logs that can't be parsed by `jq`). The function you mention is another story.

PR #400, opened 2018-06-13
Spelling corrections and fixed stack compose formatting in example

jdgiotta:
Not a big change, but contributing nonetheless.


michaelshobbs:
thanks!

PR #401, opened 2018-06-19
Use log timestamps from docker logs.

localghost:
The rationale for this PR is that not preserving original timestamps by logspout makes investigation of issues across multiple services hard.

### DIFF
--- a/router/pump.go
+++ b/router/pump.go
@@ -212,7 +212,7 @@ func (p *LogsPump) pumpLogs(event *docker.APIEvents, backlog bool, inactivityTim
 
 	// RawTerminal with container Tty=false injects binary headers into
 	// the log stream that show up as garbage unicode characters
-	rawTerminal := false 
+	rawTerminal := false
 	if allowTTY && container.Config.Tty {
 		rawTerminal = true
 	}
@@ -235,6 +235,7 @@ func (p *LogsPump) pumpLogs(event *docker.APIEvents, backlog bool, inactivityTim
 				Since:             sinceTime.Unix(),
 				InactivityTimeout: inactivityTimeout,
 				RawTerminal:       rawTerminal,
+				Timestamps:        true,
 			})
 			if err != nil {
 				debug("pump.pumpLogs():", id, "stopped with error:", err)
@@ -361,10 +362,15 @@ func newContainerPump(container *docker.Container, stdout, stderr io.Reader) *co
 				}
 				return
 			}
+			logMessage, logTime, err := parseLogLine(line)
+			if err != nil {
+				debug("pump.newContainerPump():", normalID(container.ID), ", failed to parse log line:", err)
+				continue
+			}
 			cp.send(&Message{
-				Data:      strings.TrimSuffix(line, "\n"),
+				Data:      logMessage,
 				Container: container,
-				Time:      time.Now(),
+				Time:      logTime,
 				Source:    source,
 			})
 		}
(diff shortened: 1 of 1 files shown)


gbolo:
Hi @localghost 
This is an interesting idea. I would imagine that the timestamps would be close since logspout retrieves the logs fairly quickly. I guess you would like to preserve the timestamp returned by the docker API. I would consider this if you can make it optional/configurable. Maybe you can introduce a bool flag, and by default we would use `time.Now()`

router/pump.go
```diff
@@ -235,6 +235,7 @@ func (p *LogsPump) pumpLogs(event *docker.APIEvents, backlog bool, inactivityTim
 				Since:             sinceTime.Unix(),
 				InactivityTimeout: inactivityTimeout,
 				RawTerminal:       rawTerminal,
+				Timestamps:        true,
```
gbolo: set this to true only if this feature is enabled

router/pump.go
```diff
@@ -361,10 +362,15 @@ func newContainerPump(container *docker.Container, stdout, stderr io.Reader) *co
 				}
 				return
 			}
+			logMessage, logTime, err := parseLogLine(line)
```
gbolo: `parseLogLine` should only be called if this feature is enabled. Also, if we get an error, perhaps we should fall back to default behaviour of using `Time.Now()`

router/pump.go
```diff
@@ -361,10 +362,15 @@ func newContainerPump(container *docker.Container, stdout, stderr io.Reader) *co
...
+			if err != nil {
+				debug("pump.newContainerPump():", normalID(container.ID), ", failed to parse log line:", err)
+				continue
+			}
 			cp.send(&Message{
-				Data:      strings.TrimSuffix(line, "\n"),
+				Data:      logMessage,
 				Container: container,
-				Time:      time.Now(),
+				Time:      logTime,
```
gbolo: again, the value of this would depend if this feature is enabled or not.

router/pump.go
```diff
@@ -396,3 +402,12 @@ func (cp *containerPump) remove(logstream chan *Message) {
...
 	delete(cp.logstreams, logstream)
 }
+
+func parseLogLine(line string) (string, time.Time, error) {
+	logEntry := strings.SplitN(strings.TrimSuffix(line, "\n"), " ", 2)
+	logTime, err := time.Parse(time.RFC3339Nano, logEntry[0])
+	if err != nil {
+		return "", time.Time{}, err
+	}
+	return logEntry[1], logTime, nil
```
gbolo: there is no guarantee that `logEntry[1]` exists. You should rework this function to check the length of `logEntry`

localghost:
@gbolo 
Thanks for review. I will try to apply necessary fixes in a few days.

router/pump.go
```diff
@@ -405,3 +411,22 @@ func (cp *containerPump) remove(logstream chan *Message) {
...
+	logEntry := strings.SplitN(line, " ", 2)
+	logTime, err := time.Parse(time.RFC3339Nano, logEntry[0])
+	if err != nil {
+		return line, time.Now()
+	}
+
+	if len(logEntry) == 2 {
+		return logEntry[1], logTime
+	}
+	return "", logTime
```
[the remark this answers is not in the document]
localghost: if `line` has no white spaces - if it had the function would finish at `len(logEntry) == 2` - and `logEntry[0]` is a valid time string - otherwise the function would finish at `err != nil` then it means that an empty message was logged

localghost:
@gbolo, I know, I just don't have time currently. I will update this pr in a few days. Thanks.

localghost:
@gbolo @michaelshobbs 
I have added tests but the CI failed, however, the failure does not look to be connected to my changes - maybe caused by recent github problems.

localghost:
@gbolo @michaelshobbs 
CI is green

yunusemrecatalcam:
It looks like you changed back it to time.Now() instead logTime. I am searching for a way to getting miliseconds of the dates. Since they are all 0, observing exceptions is hard. Is there a way to do it in updated version?  
Screenshot_2019-10-22 Graylog - Search

arseniybanayev:
I have the same problem as @yunusemrecatalcam. What is the workaround for seeing real timestamps with at least millisecond resolution?

PR #406, opened 2018-07-11
Update Alpine to 3.8

develar:


michaelshobbs:
thanks for the PR! 

looks like `Dockerfile.dev` needs to be updated as well
https://github.com/gliderlabs/logspout/blob/fefb70f20ceddba4543b5b7b53a4ed2b1ede1697/Dockerfile.dev#L1

PR #408, opened 2018-08-06
Adding TLS client auth and more control of CA trust store

gbolo:
Thought I would extend TLS support in logspout for more advanced usage. Since I currently work for a security company, I am quite familiar with TLS (even read some RFCs) and also have some experience in writing TLS implementations in other Go projects. I have reviewed the other PRs which addressed this issue and hope that this PR addresses all concerns that the authors of this application had with them.

I have added many comments in the code to explain what I was thinking, not sure if that is desirable or not. Also note that due to the explicit use of EC curve `X25519` you will need `go 1.8+` to compile. If this is not desirable, I can remove this. We can also enforce this by adding `// +build go1.8`

In terms of test cases, there are many I have created. Ofcourse there can be many more. Considering that there were none, I feel that I added enough. Also note that the test cases only test that the manipulation of `tls.Config` is correct. I feel that it is unnecessary to actually test a TLS connection, as this would mean that we would be duplicating test case work that should already be present in the Go standard library which is beyond the scope of this project :)

Also, I will do one more commit to amend the existing documentation to include the new TLS settings.


gbolo:
documentation has been added

gbolo:
ping @josegonzalez @michaelshobbs @mattatcha 

Sorry for the bother, but can we begin the review process for this PR? I know I may sound impatient, but I am working on a project that may incorporate this and do not wish to maintain a fork for very long. If you are all too busy, perhaps I can ask a colleague of mine (who happens to know TLS very well) to review it. Please advise.

Thanks for your time.

transports/tls/tls.go
```diff
@@ -22,9 +53,101 @@ func rawTLSAdapter(route *router.Route) (router.LogAdapter, error) {
...
+
+// createTLSConfig creates the required TLS configuration that we need to establish a TLS connection
+func createTLSConfig() (*tls.Config, error) {
+
+	var err error
+	tlsConfig := &tls.Config{}
+
+	// use stronger TLS settings if enabled
+	// TODO: perhaps this should be default setting
+	if os.Getenv(envTLSHardening) != "true" {
```
michaelshobbs: please forgive my naivety but what are these set to if `envTLSHardening` is `true`?

gbolo: Nice catch, the condition should have read `==`. I committed the fix

gbolo:
@michaelshobbs yes I am running gofmt from my editor. Do you need me to remove new lines and run gofmt with any specific flags?

transports/tls/tls.go
michaelshobbs: Can you make sure the tests cover both configurations?

michaelshobbs:
Hrm. You have some extra new lines after the beginning and before the end of functions. Perhaps `gofmt` doesn’t handle this. In any case, those are unnecessary.

gbolo:
@michaelshobbs I have added a new test case for hardening and also removed extra new lines at the beginning and end of functions. I also confirmed that `gofmt -s` does not care about these extra lines.

gbolo:
Hi @michaelshobbs, do you require anything else? If you guys need some help maintaining this project, I can provide some assistance. In any case, let's get this reviewed and merged.

Thanks.

michaelshobbs:
> do you require anything else?

Just more time in any given day 😄We could definitely use some help with this project. Any time you can spare would be greatly appreciated.

transports/tls/tls.go
```diff
@@ -1,30 +1,162 @@
...
+
 func init() {
 	router.AdapterTransports.Register(new(tlsTransport), "tls")
 	// convenience adapters around raw adapter
 	router.AdapterFactories.Register(rawTLSAdapter, "tls")
+
+	// we should load our TLS configuration only once
+	// since it is not expected to change during runtime
+	var err error
+	clientTLSConfig, err = createTLSConfig()
```
michaelshobbs: i think a more concise syntax would be 
```
if clientTLSConfig, err := createTLSConfig(); err != nil {
    log.Fatalf("error with TLSConfig: %s", err)
}

File: transports/tls/tls.go [hunk 3]
```diff
@@ -1,30 +1,162 @@
...
 }
 
-type tlsTransport int
-
 func (t *tlsTransport) Dial(addr string, options map[string]string) (net.Conn, error) {
-	conn, err := tls.Dial("tcp", addr, nil)
+	// at this point, if our trust store is empty, there is no point of continuing
+	// since it would be impossible to successfully validate any x509 server certificates
+	if len(clientTLSConfig.RootCAs.Subjects()) < 1 {
+		return nil, fmt.Errorf("FATAL: TLS CA trust store is empty! Can not trust any TLS endpoints: tls://%s", addr)
```
michaelshobbs: i generally prefer named return vars such that this would become:
```
    err = fmt.Errorf("FATAL: TLS CA trust store is empty! Can not trust any TLS endpoints: tls://%s", addr)
    return
```

transports/tls/tls.go
gbolo: yea i originally tried that, but since `clientTLSConfig` is a variable in the package scope, you get the following error when compiling:

```
# github.com/gliderlabs/logspout/transports/tls
transports/tls/tls.go:69:26: clientTLSConfig declared and not used
```

so that's why i declared the `err` then used `=` instead of `:=`

transports/tls/tls.go
gbolo: OK no problem. Chaning to:

```
if len(clientTLSConfig.RootCAs.Subjects()) < 1 {
	err := fmt.Errorf("FATAL: TLS CA trust store is empty! Can not trust any TLS endpoints: tls://%s", addr)
	return nil, err
}
```

michaelshobbs: ah i actually meant to say perhaps we should use the following style throughout this file (and subsequently the project)
```
func (t *tlsTransport) Dial(addr string, options map[string]string) (conn net.Conn, err error) {
  if len(clientTLSConfig.RootCAs.Subjects()) < 1 {
    err = fmt.Errorf("FATAL: TLS CA trust store is empty! Can not trust any TLS endpoints: tls://%s", addr)
    return
  }
  // attempt to establish the TLS connection
  if conn, err = tls.Dial("tcp", addr, clientTLSConfig); err != nil{
    return
  }
  return
}
```
note the named return variables in the function definition and the use of `return` without args. thoughts?

transports/tls/tls.go
michaelshobbs: ah derp yea.... ok so then it could become:
```
	var err error
	if clientTLSConfig, err = createTLSConfig(); err != nil {
		// without a valid/desired TLS config, we should exit
		log.Fatalf("error with TLSConfig: %s", err)
	}
```

gbolo:
Hi @michaelshobbs,
I agree that we should all strive to have a consistent style within a project. I have updated the PR to include the desired style changes. I have also looked through the commit log since last release and have updated the `CHANGELOG.md` file.

I would not mind helping out (when I can) to maintain this project. If there's a slack chat, or something similar, that you and the other maintainers frequent; please let me know. Also, I may need some elevated privileges in this repo in order to be more effective (perhaps write access?).

Thanks for merging this.

transports/tls/tls.go
gbolo: @michaelshobbs that's how I prefer it as well. The reason I didn't name my return values was because I noticed that the rest of the project does not and I was trying to be consistent. I will change it to that style now.

transports/tls/tls.go
```diff
@@ -1,30 +1,161 @@
...
 
-type tlsTransport int
-
 func (t *tlsTransport) Dial(addr string, options map[string]string) (net.Conn, error) {
-	conn, err := tls.Dial("tcp", addr, nil)
+	// at this point, if our trust store is empty, there is no point of continuing
+	// since it would be impossible to successfully validate any x509 server certificates
+	if len(clientTLSConfig.RootCAs.Subjects()) < 1 {
+		err := fmt.Errorf("FATAL: TLS CA trust store is empty! Can not trust any TLS endpoints: tls://%s", addr)
+		return nil, err
```
michaelshobbs: just saw your comment on naked vs named returns. since we're using naked returns here, this can be reverted to it's previous state.

```
return nil, fmt.Errorf("FATAL: TLS CA trust store is empty! Can not trust any TLS endpoints: tls://%s", addr)
```

apologies for the churn here and thanks for engaging in the style discussion

gbolo: OK, I have rewritten all functions to use naked returns

PR #411, opened 2018-08-17
Add bind address option for routesapi

chopmann:
Did not see a reason why not to bind the HTTP Server to an specific Address. I assume most deployments just run the container with one interface, but if not or if you just run the binary outside of docker, you might want to pin the listening address. 

The default behavior has not changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,7 @@ If you use multiline logging with raw, it's recommended to json encode the Data 
 * `DEBUG` - emit debug logs
 * `EXCLUDE_LABEL` - exclude logs with a given label
 * `INACTIVITY_TIMEOUT` - detect hang in Docker API (default 0)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -169,6 +169,7 @@ If you use multiline logging with raw, it's recommended to json encode the Data
 * `DEBUG` - emit debug logs
 * `EXCLUDE_LABEL` - exclude logs with a given label
 * `INACTIVITY_TIMEOUT` - detect hang in Docker API (default 0)
+* `HTTP_BIND_ADDRESS` - configure which address to listen on (default 0.0.0.0)
```
gbolo: please change to:
`HTTP_BIND_ADDRESS` - configure which address/interface to listen on (default 0.0.0.0)

gbolo:
Also, please add your change to the `CHANGELOG.md` file

gbolo:
This has been squashed and merged. Thanks for your contribution

PR #416, opened 2018-09-12
Added ability to skip containers 

ibrokethecloud:
Currently EXCLUDE_LABEL expects the containers to have labels of the form
`label_name:true` 

This works fine for certain scenarios but cant handle scenarios in kubernetes where we want to ignore kube-system containers.

The PR allows EXCLUDE_LABELS to be of the form `label_name` or `label_name:value`

This should keep it backward compatible while allowing us to use new label key value pairs to handle cases where we can ignore entire namespaces in k8s.

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,7 +167,7 @@ If you use multiline logging with raw, it's recommended to json encode the Data 
 * `BACKLOG` - suppress container tail backlog
 * `TAIL` - specify the number of lines in the log tail to capture when logspout starts (default `all`)
 * `DEBUG` - emit debug logs
-* `EXCLUDE_LABEL` - exclude logs with a given label
+* `EXCLUDE_LABEL` - exclude logs with a given label. The variable can be of the form `label_name` or `label_name:label_value`. If no explicit key:value pair is specified then the default label matching will be of the form `label_name = true`.
 * `INACTIVITY_TIMEOUT` - detect hang in Docker API (default 0)
 * `HTTP_BIND_ADDRESS` - configure which interface address to listen on (default 0.0.0.0)
 * `PORT` or `HTTP_PORT` - configure which port to listen on (default 80)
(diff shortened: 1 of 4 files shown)


gbolo:
Hi @ibrokethecloud,

This is a good idea, I will review this shortly

router/pump.go
```diff
@@ -86,9 +86,29 @@ func ignoreContainer(container *docker.Container) bool {
 			return true
 		}
 	}
-	excludeLabel := getopt("EXCLUDE_LABEL", "")
+
```
gbolo: all changes to this file's logic can be simplified like:

```
excludeString := getopt("EXCLUDE_LABEL", "")
excludeValue := "true"
// support EXCLUDE_LABEL having a custom label value
if len(strings.Split(excludeString, ":")) == 2 {
  excludeValue = excludeStringArr[1]
}

if value, ok := container.Config.Labels[excludeLabel]; ok {
  return len(excludeLabel) > 0 && strings.ToLower(value) == strings.ToLower(excludeValue)
}
return false
```

ibrokethecloud: thanks for the feedback.. i have updated the PR.

gbolo: > thanks for the feedback.. i have updated the PR.

the update i see only added comments... https://github.com/ibrokethecloud/logspout/commit/2184712c11a6a1c0895afbc77effba03361a3de2

README.md
```diff
@@ -167,7 +167,7 @@ If you use multiline logging with raw, it's recommended to json encode the Data
 * `BACKLOG` - suppress container tail backlog
 * `TAIL` - specify the number of lines in the log tail to capture when logspout starts (default `all`)
 * `DEBUG` - emit debug logs
-* `EXCLUDE_LABEL` - exclude logs with a given label
+* `EXCLUDE_LABEL` - exclude logs with a given label. The variable can be of the form `label_name` or `label_name:label_value`. If no explicit key:value pair is specified then the default label matching will be of the form `label_name = true`.
```
gbolo: I think the following explanation is easier to understand:
* `EXCLUDE_LABEL` - exclude containers with a given label. The label can have a value of `true` or a custom value matched with `:` after the label name like `label_name:label_value`

gbolo:
@ibrokethecloud thanks again for your contribution!

@michaelshobbs this change looks good to me. If you have no objections, I will merge this tomorrow.

Issue #420, opened 2018-10-15
logspout not sending logs >16k with docker 18.06 or later


nonspecialist:
It appears that some of the changes in Docker to manage large log messages have resulted in logspout not passing them on. In previous versions, the log stream always ended with a newline but since 18.06 (and possibly slightly earlier) this isn't the case.

As a result, the `bufio.ReadString()` in `router/pump.go` gets the EOF and throws away the message. For large logs, this could be the entire message if it's not newline-terminated (as some Java stack traces are).

PR #421, opened 2018-10-15
Pass on data received before EOF, which is not terminated by a newline

nonspecialist:
Addresses #420 

If we go receive an EOF, make sure we pass on any pending data before finishing up


nonspecialist:
FYI, we've been running this across about 80 worker nodes for the past 2 weeks, and it's saved quite a few log entries that would otherwise have gone missing.

mattayes (approved):
lgtm!

PR #427, opened 2018-11-12
Add arm32v6 compatible build

nourspace:
I use this arm image to collect logs from docker containers running on Raspberry PI 3+ and thought it could be useful for others.

I think we need to update README to mention that. Also, I didn't find the place responsible for pushing the final image to Docker Hub.

Suggestions are welcome.


nourchawich:
Having issue building the image on CircleCI. On my mac machine, this works without issues.

```
make build-arm
```

nourchawich:
Hey @gbolo 
thank you for the review. I actually started by doing what you suggested then got stuck on how to achieve that yet allow the arm image to be customised like the main image does.

Using multi-stage building will hide the builder image and thefore not allow users to do something like

```
FROM gliderlabs/logspout:v3.2.6-arm
```
As this image would be based on the final runtime arm image.

Do you have any thoughts on this?

gbolo:
Hey @nourchawich 
after looking through the build script. nothing in there is important to remain in the final container other than `ca-certificates` and ofcourse, the logspout binary.

I would support the use of multistage build (the regular version should also be switched to this) with an x86_64 builder image that can cross compile logspout and a final arm image that receives no RUN commands. Everything (including `ca-certificates`) can be copied in from builder image.

Do you need any help to accomplish this?

michaelshobbs:
closed in favor of #514

PR #443, opened 2019-08-17
Support swarm service labels when using filter.labels

erbrecht:
If a container was started by a swarm service, this additionally checks the service labels in the MatchContainer function. It also looks at service labels when deciding whether to ignore a container based on labels. This approach doesn't require any special configuration, it should just work seamlessly whether containers are started individually or as part of a service or stack.


erbrecht:
After submitting this, I realized that the `IgnoreContainer` function in `pump.go` should probably accept a `*docker.Container` and `*swarm.Service` instead of `*docker.Container` and `map[string]string` of labels. I think it makes more sense for them to be similar types. I made these changes in my repo and they pass the tests but I haven't pushed them yet. I wanted to see if there was any feedback on this first.

PR #459, opened 2020-03-20
Add support for multiple exclusion labels

### LINKED ISSUE #458: Add ability to specify different exclusion labels
Currently, the logspout container only allows the `EXCLUDE_LABEL` environment variable to be of the form `label_name` or `label_name:value`.

A follow-up step for this, would be to support a list of `label_name:value` separated by an divider, so that containers matching at least one of the listed labels are ignored by logspout. 

At our company, we are already deploying a modified version of logspout that allows to specify the following:

    EXCLUDE_LABEL="<label_name1:value1>;<label_name2:value2"

renehernandez:
Fix #458 

Divider between labels: ;


router/pump.go
```diff
@@ -91,15 +91,22 @@ func ignoreContainer(container *docker.Container) bool {
 
 	excludeLabel := cfg.GetEnvDefault("EXCLUDE_LABEL", "")
```
renehernandez: For backwards compatibility, I didn't change the environment variable name. Although with this change, it would make more sense to be called `EXCLUDE_LABELS`

renehernandez: Done

renehernandez:
> looks good. one suggestion inline and can you also include documentation updates?

Documentation updated

michaelshobbs:
Thanks for the contribution 🎉

renehernandez:
> Thanks for the contribution 🎉

My pleasure. Looking forward to contribute more!!

PR #472, opened 2020-05-01
Add optional TCP framing to syslog adapter

michaelshobbs:
closes #429

### DIFF
--- a/adapters/syslog/syslog.go
+++ b/adapters/syslog/syslog.go
@@ -29,11 +29,14 @@ var (
 	econnResetErrStr string
 )
 
+// TCPFraming represents the type of framing to use for syslog messages
 type TCPFraming string
 
 const (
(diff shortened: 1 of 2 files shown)


hhromic:
Hello!
Thanks for implementing this. There is no documentation at the moment for this new feature, however from the source code I understand that there is now a new `SYSLOG_TCP_FRAMING` environment variable that when set to:

* `traditional` (default) the Syslog message remains unchanged (current behaviour)
* `octet-counted` the Syslog message gets prepended with its octet size (new behaviour)

Is the above correct? If so I could help adding this information to the README file.

Cheers!

michaelshobbs:
That is correct. A `README` PR would be greatly appreciated. 🙏

hhromic:
Done, let me know if you want any changes ;)

michaelshobbs:
#429

PR #476, opened 2020-05-05
Add Syslog TCP framing documentation to README

hhromic:
This documents the new TCP framing option introduced in #472


michaelshobbs:
Thanks!

PR #485, opened 2020-08-18
fix module path

KateGo520:
Fix [#475](https://github.com/gliderlabs/logspout/issues/475)


michaelshobbs:
Thanks for the PR! Looks like linting failed here

```
logspout.go:3: File is not `goimports`-ed (goimports)
```

Issue #412, opened 2018-08-28
Panic accessing /logs via websocket


smlx:
First, thanks for the great software :-)

I'm seeing a panic when accessing /logs via a websocket. Here's the stack trace:

```
# logspout v3.2.5 by gliderlabs
# adapters: raw syslog multiline tcp udp tls
# options : persist:/mnt/routes
# jobs    : pump routes http[routes,logs]:80
# routes  : none
panic: send on closed channel

goroutine 41 [running]:
github.com/gliderlabs/logspout/router.(*containerPump).send(0xc42022f560, 0xc420053280)
	/go/src/github.com/gliderlabs/logspout/router/pump.go:384 +0x129
github.com/gliderlabs/logspout/router.newContainerPump.func1(0x5634dbc5087f, 0x6, 0x5634dc174ac0, 0xc42023c060)
	/go/src/github.com/gliderlabs/logspout/router/pump.go:364 +0xbc
created by github.com/gliderlabs/logspout/router.newContainerPump
	/go/src/github.com/gliderlabs/logspout/router/pump.go:373 +0x174

```

You can reproduce this by running logspout via docker:

    docker run -d --name="logspout" --volume=/var/run/docker.sock:/var/run/docker.sock --publish=8000:80 --restart=always gliderlabs/logspout

and then hitting the endpoint with curl:

    $ curl -H "Connection: Upgrade" -H "Upgrade: websocket" http://127.0.0.1:8000/logs
    mismatch challenge/response

`mismatch challenge/response` is the only output.

vtereso:
I also experienced this when attempting to use `gorilla/websocket` to connect to the endpoint, but when I used the `golang.org/x/net/websocket` instead, it worked as intended. Not sure what other libraries will have conflicts as well. Further, I am receiving the data from the websocket with random timestamps inserted arbitrarily inside the json response (when viewing it as a string), so it is failing to unmarshal into a route.Message struct (as internal to logspout).

semk:
logspout is expecting a browser client to access the websocket endpoint. Due to missing origin header the panic was occuring. I've added a patch which allows non-browser clients to access this endpoint. https://github.com/gliderlabs/logspout/pull/502

PR #502, opened 2021-02-23
Allow non-browser websocket clients to access the /logs endpoint

semk:
` websocket.Handler.ServeHTTP` serves to browser requests and was also causing logspout to crash when accessed by a client like `wscat` as they don't typically set the origin header (See https://github.com/gliderlabs/logspout/issues/412). This change bypasses the validation for origin header.


michaelshobbs:
Thanks for the contribution. Thoughts on a unit test or two for this and the package in general?

6r17:
Hei ; the error is sill happening and non-browser client seem to have a problem using the ws endpoint

PR #505, opened 2021-04-22
Release docker image for arm64

odidev:
The following files have been modified:

- In circle.yml file added a new job **publish** in order to use buildx to build and push the image for both the platforms to dockerhub.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -36,3 +36,19 @@ jobs:
             if [ "${CIRCLE_BRANCH}" == "release" ]; then
               make release
             fi
+  publish:
+    machine:
+      image: ubuntu-1604:202007-01
+    working_directory: /home/circleci/logspout
+    environment:
+      DEBUG: true
+    steps:
+      - checkout
+      - run: docker login -u "$DOCKER_USERNAME" -p "$DOCKER_PASSWORD"
+      - run: make publish
+
+  version: 2
+  build_and_publish:
+    jobs:
+      - build
+      - publish
--- a/Makefile
+++ b/Makefile
@@ -127,5 +127,18 @@ clean:
 	docker rm $(shell docker ps -aq) || true
 	docker rmi $(NAME):dev $(NAME):$(VERSION) || true
 	docker rmi $(shell docker images -f 'dangling=true' -q) || true
+publish:
+	mkdir -vp ~/.docker/cli-plugins/
+	curl --silent -L --output ~/.docker/cli-plugins/docker-buildx https://github.com/docker/buildx/releases/download/v0.3.1/buildx-v0.3.1.linux-amd64
+	chmod a+x ~/.docker/cli-plugins/docker-buildx
+	docker run -it --rm --privileged tonistiigi/binfmt --install all
+	docker buildx create --use --name mybuilder
+ifeq ($(CIRCLE_BRANCH), master)
+	docker buildx build --platform linux/arm64,linux/amd64 -t ${DOCKER_USERNAME}/logspout:${CIRCLE_BRANCH} -t ${DOCKER_USERNAME}/logspout:latest --push .
+endif
+
+ifeq ($(CIRCLE_BRANCH), release)
+	docker buildx build --push --platform linux/arm64,linux/amd64 -t ${DOCKER_USERNAME}/logspout:${VERSION} .
+endif
 
 .PHONY: release clean


circle.yml
```diff
@@ -36,3 +36,25 @@ jobs:
...
+      image: ubuntu-1604:202007-01
+    working_directory: /home/circleci/logspout
+    environment:
+      DEBUG: true
+    steps:
+      - checkout
+      - run: mkdir -vp ~/.docker/cli-plugins/
+      - run: curl --silent -L --output ~/.docker/cli-plugins/docker-buildx https://github.com/docker/buildx/releases/download/v0.3.1/buildx-v0.3.1.linux-amd64
+      - run: chmod a+x ~/.docker/cli-plugins/docker-buildx
+      - run: docker login -u "$DOCKER_USERNAME" -p "$DOCKER_PASSWORD"
```
michaelshobbs, line 50: how are `DOCKER_USERNAME` and `DOCKER_PASSWORD` defined?

circle.yml
```diff
@@ -36,3 +36,25 @@ jobs:
...
+      - run: docker run -it --rm --privileged tonistiigi/binfmt --install all
+      - run: docker buildx create --use --name mybuilder
+      - run: |
+          docker buildx build --push --platform linux/arm64,linux/amd64 -t ${DOCKER_USERNAME}/logspout:${CIRCLE_BRANCH} -t ${DOCKER_USERNAME}/logspout:latest .
```
michaelshobbs, line 54: doesn't this overwrite `latest` every time this job is run? we would only want latest to be overwritten on the `master` branch

michaelshobbs (changes requested):
if we're going to build multiple architectures on branches then we also need to implement a plan to do the same as part of the release cycle. anybody willing to pick that work up?

circle.yml
odidev: To set **DOCKER_USERNAME** and **DOCKER_PASSWORD**, we need to choose the **Environment Variables** option and set the username and password for the dockerhub in **CircleCI project settings**.

circle.yml
odidev: Please review, I have done the changes as per your suggestion.

michaelshobbs: we also need a case for the `release` branch. you can get the release version from the `VERSION` file at the repo root

circle.yml
```diff
@@ -36,3 +36,28 @@ jobs:
...
+      - run: mkdir -vp ~/.docker/cli-plugins/
+      - run: curl --silent -L --output ~/.docker/cli-plugins/docker-buildx https://github.com/docker/buildx/releases/download/v0.3.1/buildx-v0.3.1.linux-amd64
+      - run: chmod a+x ~/.docker/cli-plugins/docker-buildx
+      - run: docker login -u "$DOCKER_USERNAME" -p "$DOCKER_PASSWORD"
+      - run: docker run -it --rm --privileged tonistiigi/binfmt --install all
+      - run: docker buildx create --use --name mybuilder
+      - run: docker buildx build --push --platform linux/arm64,linux/amd64 -t ${DOCKER_USERNAME}/logspout:${CIRCLE_BRANCH} .
+      - run: |
+            if [ "$CIRCLE_BRANCH" = "master" ]; then
+               docker buildx build --push --platform linux/arm64,linux/amd64 -t ${DOCKER_USERNAME}/logspout:latest .
```
michaelshobbs, line 56: i'm not familiar with buildx. does this end up pushing two different images? if so, what are their names/tags?

circle.yml
odidev: I have done the changes as per your suggestion. Please let me know if any more changes are required.

circle.yml
odidev: When it will run on master branch one image with two different tags will be published 

1) gliderlabs/logspout:master 

2) gliderlabs/logspout:latest 

When it will run on release branch, image will be publish in dockerhub as gliderlabs/logspout:v3.2.15

michaelshobbs: That makes sense. However my question was about the two architectures. Do we wind up with one for `linux/arm64` and one for `linux/amd64`? As an end-user, how would I get the `linux/arm64` version for example?

michaelshobbs:
Looks like there is a yaml error according to CircleCI

circle.yml
michaelshobbs: Looks like docker pulls the correct image for your target architecture automatically

michaelshobbs:
I merged these changes in #512

michaelshobbs:
Thank you for tackling this! 🙏

michaelshobbs:
Looks good on master

Screen Shot 2021-12-03 at 13 59 16

